PR #34, opened 2020-12-27
Platformio support

nils-werner:
This PR provides platformio support by contributing/fixing the following things:

1. It removes the `libaries/ArduinoOTA` dir. `ArduinoOTA` is part of `FrameworkArduino` and doesn't have to be included in this repo. On the contrary: if the included library is incompatible with the installed `FrameworkArduino`, [compiler errors appear](https://forum.blueforcer.de/d/394-meldung-u-spiffs-was-not-declared)
 1. It moves the sources into their own directory, `rancilio-pid/src/`, which is not a parent dir of `rancilio-pid/libraries/`. My compiler didn't like the two to be inside each other and was throwing import errors.
 1. A very simple config file for platformio.

With this PR, the firmware can be compiled and programmed using

```
pip install platformio
pio run --target=upload
```


btotherunner (commented):
thx!

btotherunner (approved):
thx!

btotherunner (approved):
thx!

btotherunner (approved):
thx!

PR #76, opened 2021-03-10
Water level measurement functional again; fix compile error if DISPLAY = 0

andreas301-sketch:
continuous water level measurement disappeared from code in master upon merge, back now
included new function in display.h to prevent compile error if DISPLAY =0


murmeltier08:
sieht gut aus. ist übernommen

PR #85, opened 2021-03-21
B multilingual support

gomezalberto:
This is my effort to have multi-lingual support. I have only bothered with the text that appears in the display. All text that is used for debugging, which was in English already, is not affected by this.

Currently added Spanish and English to German. So far seems to work. But I have only tested with PID only + brewswitch.

### DIFF
--- a/rancilio-pid/userConfig.h
+++ b/rancilio-pid/userConfig.h
@@ -20,7 +20,7 @@
 #define HEATINGLOGO 0              // 0 = deactivated, 1 = Rancilio, 2 = Gaggia 
 #define OFFLINEGLOGO 1             // 0 = deactivated, 1 = activated
 #define BREWSWITCHDELAY 3000       // time in ms
-#define LANGUAGE 2                 // 0 = DE, 1 = EN, 2 = ES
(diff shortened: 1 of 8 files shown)


murmeltier08:
Thank you for your Input, i will check it soon!

rancilio-pid/userConfig.h
```diff
@@ -49,8 +50,10 @@
 
 /// Wifi 
 #define HOSTNAME "Rancilio"
-#define D_SSID "myssid"
-#define PASS "mypass"
+//#define D_SSID "myssid"
+//#define PASS "mypass"
+#define D_SSID "D45C Hyperoptic 1Gb Fibre 2.4Ghz"
```
gomezalberto: Apologies -these are the values from my dad's wifi. Can be safely ignored

rancilio-pid/userConfig.h
```diff
@@ -68,8 +71,10 @@
 #define MQTT_SERVER_PORT 1883    
 
 // BLynk
-#define AUTH "myauth"
```
gomezalberto: Also apologies for including these blynk details -can be safely ignored here too.

PR #100, opened 2021-04-11
Change folder structure to support both ArduinoIDE and PlatformIO in Visual Studio Code

mikulcak:
Hi,

I have moved the source files in `rancilio-pid` to a new subfolder, also called `rancilio-pid` and edited `platformio.ino` accordingly. This way, ArduinoIDE and Visual Studio Code / PlatformIO can both be used.


mikulcak:
Oh, and I also removed and ignored .DS_Store files because I couldn't clone the repository on my MacBook, I only got:

```
error: Untracked working tree file '.DS_Store' would be overwritten by merge.
fatal: unable to checkout working tree
warning: Clone succeeded, but checkout failed.
```

That's fixed now.

PR #104, opened 2021-04-16
Not compiling when DISPLAY == 0

lebuni:
If ´DISPLAY==0´, there is a closing brace without an opening one.


murmeltier08:
Thank you!

PR #110, opened 2021-04-24
Move bPID.Compute from ISR to loop

lebuni:
The ESP32 can't execute non-IRAM functions inside an ISR. It may cause the _Guru Meditation Error: Core  1 panic'ed (Cache disabled but cached memory region accessed)_. 
Therefore I would suggest to move `bPID.Compute()` from the timer ISR to the loop. `bPID.Compute()` will only change the output in intervals of windowSize = 1000ms, so the precise timing provided by the timer interrupt is not really necessary.

### DIFF
--- a/rancilio-pid/rancilio-pid/ISR.h
+++ b/rancilio-pid/rancilio-pid/ISR.h
@@ -1,54 +1,38 @@
 /********************************************************
-    Timer 1 - ISR for PID calculation and heat realay output
+Timer 1 - ISR for PID calculation and heat realay output
 ******************************************************/
 
+#ifdef ESP8266
+	void ICACHE_RAM_ATTR onTimer1ISR() {
+		timer1_write(6250); // set interrupt time to 20ms
 
+		if (Output <= isrCounter) {
+			digitalWrite(pinRelayHeater, LOW);
+		} else {
+			digitalWrite(pinRelayHeater, HIGH);
+		}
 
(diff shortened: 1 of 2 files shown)


Miau09:
this can (must) be fixed by simply applying the "ICACHE_RAM_ATTR" or "IRAM_ATTR" to the PID-library compute() function.

It was already published for the ESP8266 in our chat, though it seems like it was not done in our github library...

I agree this is not the best solution because updates of the library will overwrite the files, but this library has not been updated for years so it seems not like an issue for me.

Bigger problem in my opinion is that the PID needs to be computed at the 1s window, delays of some hundred ms can cause unwanted behavior in the PID output.
I do not really want to push a change like this without testing, especially on the 8266 with the high number of users.

The ESP32 is a different story, this is still in development and will be changed later anyway because we want to implement free RTOS and use both cores for different tasks.

Did you do some testing if this changes the PID behaviour?

lebuni:
My Gaggia runs with the bPID.Compute outside the ISR for about three month now and there was no need to change PID parameters. A benefit is that OTA now works on the first try every time, that was not the case before.

My minimum output is always about 40ms when idling. So the loop has about 40ms to execute bPID.Compute, before the new output is needed. Worst case would be using the same output for two seconds in a row. But usually the temperature (and the corresponding output) isn't that much different from one second to the other from my experience.

murmeltier08:
Will done in the next Version.

PR #123, opened 2021-05-30
debounce 1.0

BlinkingJoe:
digitalRead PVS only once in brewdetection()


FunnyGermanGuy:
There ist already a debounce funtion in the normal brewswitch. Maybe a better way would be to make the function golbally available in the project for all kinds of switches rather than programming a debounce for each individually

BlinkingJoe:
What I did is not a real debouncing. I just changed two subsequent digitalRead commands to only one. Sort of poor man debouncing in that special context of code. A genuine debouncing can be added any time later if nececessary. At the moment I don't expect that to be necessary.

PR #132, opened 2021-09-17
Fix 100% duty cycle

danieltwagner:
In the current state when the PID controller requests a 100% duty cycle we will disable the heater for one timer interrupt cycle before re-enabling it. This special case allows us to keep the heater on 100% of the time

### DIFF
--- a/rancilio-pid/rancilio-pid/ISR.h
+++ b/rancilio-pid/rancilio-pid/ISR.h
@@ -13,8 +13,7 @@
     void ICACHE_RAM_ATTR onTimer1ISR() {
     timer1_write(6250); // set interrupt time to 20ms
 
-    if ((Output <= isrCounter) && (Output < windowSize)) {
-        // turn off for this window (only if PID doesn't desire 100% duty cycle)
+    if (Output <= isrCounter) {
         digitalWrite(pinRelayHeater, LOW);
     } else {
         digitalWrite(pinRelayHeater, HIGH);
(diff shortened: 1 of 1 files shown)


danieltwagner:
It occurred to me that the reason for the single off-cycle was that the ISR was running one too many times per window. This new fix also makes PID output 981 behave identical to 1000 as intended.

Example, assuming interrupts every 20ms on nodemcu and 1000ms window size, PID output = 981:
- ISR runs with isrCounter = 0, processing changes due 0ms after window start: heater goes on, isrCounter is now 20
- ISR runs with isrCounter = 20, processing changes due 20ms after window start: heater remains on, isrCounter is now 40
- ...
- ISR runs with isrCounter = 980, processing changes due 980ms after window start: heater remains on, isrCounter is now 1000

Before the next steps would have been:
- ISR runs with isrCounter = 1000, processing changes due 1000ms after window start: heater turns off (!), isrCounter is now 0
Notice that we're adding an extra ISR execution after 1000ms have elapsed in the window, when really that is the same as when 0 ms have elapsed. The heater should never turn off.

With the change we instead start from the top (isrCounter = 0)

danieltwagner:
Posting for visibility what I just mentioned in chat: The PR implies that what before was 1/51 ISR cycles will become 1/50 and so on, meaning every PID setting is going to get about 2% "hotter". Arguably this could therefore be considered a breaking change. I'm personally not sure if this will require adjusting PID parameters as I haven't finished my installation yet and hence cannot test.

PR #134, opened 2021-09-22
Return result from non-void function

danieltwagner:
As per description. Alternatively we could turn this function into a void as the result isn't ever used (obviously).
While I was in there I also changed the conditional into a preprocessor directive.


Miau09 (approved):
fine for me.
At some point we need to review the code to use preprocessor directive on everything that is not used, includes the mqtt_publish() calls in the code, etc.

PR #159, opened 2022-01-04
add eeprom write callback

danieltwagner:
open for comments :)


danieltwagner:
just commenting here to say that this doesn't actually persist yet. not quite sure why. the call returns 1, indicating an error.

PR #175, opened 2022-01-11
Platformio sample file

lauuuuuuurah:
Added Platformio Sample file. 
Users need to copy file to platformio.ini and should add own credentials.


btotherunner (approved):

btotherunner:
Passt so :D

PR #188, opened 2022-01-25
Move some hardware-related settings out of userConfig.h

fiendie:
Put common and platform-specific pin mappings in their own header files


fiendie:
Ich mach den mal zu, ist schon etwas stale :)

PR #199, opened 2022-02-04
overworked firmware version check

endym:
- comparing strings at preprocessor level is not supported
- userConfig "match check" does not include the branch name (see 1st point)
- use separate defines for all firmware version parts to be more flexible
- added a general function to get the version as string (w/o any other stuff)
  this function can be used to build further strings containing the firmware version
- introduced rancilio-pid header file so that other modules may call some functions

### DIFF
--- a/src/rancilio-pid.cpp
+++ b/src/rancilio-pid.cpp
@@ -7,13 +7,15 @@
  */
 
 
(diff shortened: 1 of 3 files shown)


endym:
Der PR soll als Vorschlag dienen.

PR #205, opened 2022-02-06
Draft: Option (TEMPSENSOR==9) to mock temperature sensor via analogREAD

phi-mah:
Apart from implementing the TEMPSENSOR 9 option, I did some cleanup w.r.t to `refreshTemp` function.

Current rebase compiles, will test with breadoard and ESP8266 this evening. Had no chance to test with ESP32 though.


kjyv:
What is this for really, why should this it be merged for everyone?

PR #209, opened 2022-02-07
initial influxdb auth

benjaminfrank:
I tested the PR on my machine and it was able to connect and write to influxdb using user and password

.

### DIFF
--- a/src/rancilio-pid.cpp
+++ b/src/rancilio-pid.cpp
@@ -167,11 +167,7 @@ unsigned int MQTTReCnctFlag;       // Blynk Reconnection Flag
 unsigned int MQTTReCnctCount = 0;  // Blynk Reconnection counter
 
 // InfluxDB Client
-#if (INFLUXDB_AUTH == 1)
-    InfluxDBClient client;
-#else
-    InfluxDBClient client(INFLUXDB_URL, INFLUXDB_DB_NAME);
-#endif
+InfluxDBClient client(INFLUXDB_URL, INFLUXDB_DB_NAME);
 Point sensor("machinestate");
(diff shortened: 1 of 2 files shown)


src/rancilio-pid.cpp
```diff
@@ -1989,6 +1989,11 @@ void setup() {
                 mqtt.setCallback(mqtt_callback);
                 checkMQTT();
             }
+
+            if(INFLUXDB_USER != "" && INFLUXDB_PASSWORD != "") {
```
endym, line 1993: String comparisons do not work this way in C/C++. Maybe better use:
`if ((strlen(INFLUXDB_USER) > 0) && (strlen(INFLUXDB_PASSWORD) > 0))`
or maybe safer (didn't tested this code):
`if ((String(INFLUXDB_USER).length() > 0) && (String(INFLUXDB_PASSWORD).length() > 0))`

benjaminfrank: done

endym (commented):
Looks good! Did you tested it with non-empty strings for `INFLUXDB_USER` and `INFLUXDB_PASSWORD`?

benjaminfrank:
yes, that is what I am running on my machine

endym (approved):
Ok, then it's ok from my side :-)

benjaminfrank:
hm it might be a good idea to check if it properly works with an empty string though

endym:
Yes, indeed.

fiendie:
@benjaminfrank I cherry-picked your commit into #218 and added API Token authentication for InfluxDB >=2.x so this PR can be closed

PR #227, opened 2022-02-27
Alpha 3.0.0/website pid values per machine state

SimonCBell:
Update sendTempEvent to sendStatusEvent 
and included machine state and PID settings for regular updates on webpage

 - PID settings taken directly from bPID.Get() and sent to status.js via JSON


kjyv:
I think this is superseded in PR #243 where parameters are received and set with a separate call

PR #228, opened 2022-02-27
Alpha 3.0.0/dallas sensor async

SimonCBell:
Run Dalla Sensor in Asynchronous Mode, so that temp read does not sit there waiting for the temp sensor to respond.

Advantage: looppid() runs at higher frequency, and there is no delay in brew switch on to actually going in brew mode


SimonCBell:
branched from wrong location...

PR #229, opened 2022-02-27
Alpha 3.0.0/dallas temp improve

SimonCBell:
Run dallas temp sensor in Asynchronous mode, so that temp read does not sit and wait for sensor to convert 

Advantage: looppid() runs faster, and there is no delay between press brew switch and actually starting brew()


SimonCBell:
wrong destination - want to merge to Alpha-3.0.0 not master

PR #230, opened 2022-02-27
set dallas temp driver to run Asyncronous, rather than sitting and wa…

SimonCBell:
Run dallas temp sensor in Asynchronous mode, so that temp read does not sit and wait for sensor to convert 

Advantage: looppid() runs faster, and there is no delay between press brew switch and actually starting brew()


murmeltier08:
Bist du bei uns im chat?

PR #231, opened 2022-02-28
Add Option to tempLED HIGH or LOW at brew/Steam ready and Correct check to tempLED at SteamON

tml89:
Add option to TEMPLED to set led pin low when brew or steam set point is within range instead of high. (Imitate Silvia deflault behavior)


murmeltier08:
Bist du bei uns im chat?

PR #232, opened 2022-03-01
Alpha 3.0.0/web param validation

SimonCBell:
Check that requested parameter update is inside allowed limits for parameter


SimonCBell:
Misunderstanding on desired feature! 

Front End validation wanted, not the here implemented Back End.

Closing without merge

PR #238, opened 2022-05-29
Add support for WiFiManager, remove soft AP

fiendie:


kYc0o:
Hi guys!

I find the additions and the work done here wonderful! However, a little explanation of what was done here would be really really awesome for the little users like me :see_no_evil: 

I'm now trying to find out through the code what really happened with this PR/commit (some splitting would have been wonderful as well!) but I'm really lost... the whole WiFi connection logic was changed, so now I don't know if my machine will connect to the WiFi configured on `userConfig.h` or it will only work on AP mode...

Well, I hope someone could describe roughly the new connection logic. Anyways thank you so much for the efforts!

fiendie:
Hi @kYc0o, 

fair point :)
The documentation will be updated eventually but the pull request is a bit sparse on the details admittedly.

The WiFiManager opens up an ad hoc WiFi network that you have to join. The hostname and passphrase are taken fron the userConfig.h, namely "HOSTNAME" and "PASS". The default passphrase is "CleverCoffee".

Once you joined the network your operating system should automatically open the capture portal where you can configure your WiFi connection, otherwise you have to manually enter the IP of the ESP into your browser.

Hope that helps!

Cheers
Andy

kYc0o:
That definitely helps a lot! Thank you!!

We may leave here a link to the documentation once it's done? I may contribute if needed.

Thanks again for the great work!

PR #243, opened 2022-07-06
Improve PID controller, parameters page and page caching

kjyv:
This PR includes multiple changes, the main changes are:

* Uses a fork of the Arduino-PID-Library (by me) with added low-pass filtering of the D part in the PID controller and allowing to limit the integrator output (added in platformio_sample.ini, needs to be added to existing platformio.ini manually).
* Allow users to set the integrator output limit from the website
* Allow users to set a setpoint offset that will increase the actual temperature that is heated to (to compensate heat loss to group)
* More accurate parameter labels (P-Tn-Tv form) on the Parameters page, additional help texts in popups and links to articles explaining PID parameters
* Improved parameters page that loads the information async and that can be cached in the browser, various performance improvements
* Show machine uptime when not brewing on standard display template (useful to know when machine is really heated through)
* Allow turning off brew detection PID and cold start PID (can be set in web interface)
* Allow setting a delay time before brew detection PID is enabled (and during which normal PID is disabled) (for why this makes sense, see: https://www.kaffee-netz.de/threads/installation-eines-temperatursensors-in-silvia-bruehgruppe.111093/#post-1453641) (in userConfig for now)
* allow PlatformIO monitoring OTA by opening a serial port and logging to it when clients are connected
* when logging, print machinestate with state names instead of numbers, prepend time of log messages
* put a lot of strings in flash to save some RAM for e.g. template processing (using F() macro)
* use LittleFS instead of deprecated SPIFFS
* some code cleanup here and there and some translation updates
* ~~use the PID parameter form instead of the P-Tv-Tn form to keep changes of the parameters separate of each other and make parameter adjustment less confusing~~

With these changes, the D (or rather Tv) parameter can be properly used to dampen overshoot and no separate brew detection PID is necessary. It can still be used but it works fine on my Silvia with BD PID turned off.
The separate cold start PID (using PonM) is also not really necessary for me (still enabled by default), it can be disabled in the web interface.
A new VERBOSE flag (userConfig) was added that, if set to true, will log the current state, PID parameters and internal values repeatedly (useful when debugging the controller). Other existing log messages might only be needed for debugging and only printing them when this flag is enabled could be added now as well.

With these changes, a new set of PID parameters should probably be used for each machine for better performance (including the Tv parameter) but previous values will work the same as before. I'm currently using P=65, I=52, D=11.5, IMax=55 with a Rancilio Silvia (V1/V2 without insulation) and get improved behaviour without overshoot.


kjyv:
As a result, the single PID is able to heat up the machine during cold start to the setpoint, after an espresso shot and also after filling a cup with hot water without considerable overshoot and ringing. The temperature is being held right away within +-0.1°C of the the setpoint (the image is a bit less exact than current state but shows the general idea)

Bildschirmfoto 2022-06-27 um 14 53 09

kjyv:
Cold start from room temperature with current state and mentioned params
Bildschirmfoto 2022-07-06 um 11 47 35

src/userConfig_sample.h
```diff
@@ -120,15 +123,17 @@ enum MACHINE {
...
+#define STEAMSETPOINT 120          // steam temperature setpoint
+#define BREWSENSITIVITY 150        // brew detection sensitivity, be careful: if too low, then there is the risk of wrong brew detection and rising temperature
+#define AGGKP 33                   // PID Kp (regular phase)
+#define AGGTN 255                  // PID Tn (regular phase)
 #define AGGTV 0                    // PID Tv (regular phase)
+#define AGGIMAX 500                // PID Integrator Max (regular phase)
 #define STARTKP 50                 // PID Kp (coldstart phase)
-#define STARTTN 150                // PID Tn (coldstart phase)
-#define STEAMKP 150                // PID kp (steam phase)
+#define STARTTN 0.5                // PID Tn (coldstart phase)
```
murmeltier08, line 135: Warum 0.5?

kjyv: Ahja, das ist nicht mehr aktuell (war für PID, nicht P-Tv-Tn). Ich ändere die aber mal auf die Werte aus dem Manual von 45 und 130

murmeltier08:
I will test the PR next week

kjyv:
Also added improved layout of parameters page (reactive to screen width). Loads parameters from /parameters endpoint, pages can all be cached now. PonM and BD PID can be enabled/disabled on Parameters page, visible fields will adapt to options chosen and features enabled in userConfig.

Bildschirmfoto 2022-08-11 um 17 14 16
Bildschirmfoto 2022-08-11 um 17 14 08

murmeltier08:
Well done @kjyv. I will merge now :)

PR #244, opened 2022-07-28
fix timeout wifi manager

murmeltier08:
FIx timeout for Portal and timeout for wlan connection


kjyv:
Seems to work, I had a machine started in offline mode this morning for the first time - before it sometimes happened that it was just stuck at the logo but didn't heat up etc. before a restart

PR #257, opened 2022-09-14
Fix: Display timer when brew is detected

joecksma:
The uptime display broke the brewtimer in the default display.
It was introduced in 16d23dd82f2f025e9702ee01b70356b670b39a55
I only fixed the check for the brewdetection


kjyv (approved):

murmeltier08:
is fixed by other PR

PR #267, opened 2022-10-18
MAX31865 RTG temp sensor support

tbaumann:
This is so far work in progress and untested.

Referring to https://github.com/rancilio-pid/ranciliopid/issues/266

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -19,6 +19,8 @@ lib_deps =
     tobiasschuerg/ESP8266 Influxdb @ 3.9.0
     git+https://github.com/me-no-dev/ESPAsyncWebServer
     git+https://github.com/tzapu/WiFiManager
+    git+https://github.com/adafruit/Adafruit_MAX31865
+    git+https://github.com/adafruit/Adafruit_BusIO
 
 [env:nodemcuv2_usb]
 platform = espressif8266 @^2.6.3
@@ -50,6 +52,7 @@ monitor_speed = 115200
 monitor_filters = esp32_exception_decoder
 debug_tool = esp-prog
 debug_init_break = tbreak setup
+
 
 [env:esp32_ota]
 platform = espressif32 @^5
--- a/src/rancilio-pid.cpp
+++ b/src/rancilio-pid.cpp
@@ -692,8 +692,9 @@ void refreshTemp() {
                     debugPrintf("Under/Over voltage");
                 }
                 Sensor3.clearFault();
+            
     #endif
-    
+     }
 
             if (machineState != kSteam) {
                 temperature -= brewTempOffset;
--- a/src/rancilio-pid.cpp
+++ b/src/rancilio-pid.cpp
@@ -693,9 +693,9 @@ void refreshTemp() {
                 }
                 Sensor3.clearFault();
     #endif
-    }
+    
 
-            if (machinestate != kSteam) {
+            if (machineState != kSteam) {
                 temperature -= brewTempOffset;
             }
 
@@ -705,7 +705,7 @@ void refreshTemp() {
                         // be read at least one time at system startup
             }
 
-            if (Brewdetection == 1) {
+            if (brewDetectionMode == 1) {
                 calculateTemperatureMovingAverage();
             } else if (!movingAverageInitialized) {                
                 movingAverageInitialized = true;
--- a/src/rancilio-pid.cpp
+++ b/src/rancilio-pid.cpp
@@ -669,10 +669,12 @@ void refreshTemp() {
             previousMillistemp = currentMillistemp;
 
     #if TEMPSENSOR == 3
-            #if MAX31865_PT100
+            #if MAX31865_RTC_TYPE == PT100
                 temperature = Sensor3.temperature(100, 430.0);
-            #else
+            #elif MAX31865_RTC_TYPE == PT1000
                 temperature = Sensor3.temperature(1000, 4300.0);
+            #else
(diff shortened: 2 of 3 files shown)


tbaumann:
Works now. Please apologize for the mDNS commit. If you want to keep it clean I can take it out of this MR.

murmeltier08:
I will test the PR soon.

tbaumann:
Actually I have some cleanup. Give me a sec and I commit it

kjyv:
Looks good but why the MDNS stuff?

tbaumann:
> Looks good but why the MDNS stuff?

I just noticed that the hostname gets advertised (By OTA) but the webinterface doesn't. This adds a service record.
But honestly it has no real business to be in this pull request. I can take it out.

In fact I am thinking of re-writing the WLAN manager section. If I get around doing it. At the moment the code is blocking at boot until WLAN is connected and it never tries any reconnect, that's totally not necessary.
There are a few ways to switch between AP and online modus automatically...
I would throw the mDNS change in there once I get around doing it.
I might also try to port Gagguino to ESP32 instead...

murmeltier08:
We change the wifi manager, after 30 seconds the offline mode starts or do you mean something else?

kjyv:
I think mDNS might not be enabled automatically with ESP32? With ESP8266 it already works fine without explicitly adding it.
You mean at boot it would be nice if it started heating right away while also trying to connect to WiFi and retrying as well if not successful in parallel? Would be nice but also might be a bit tricky to do things in parallel?

tbaumann:
> We change the wifi manager, after 30 seconds the offline mode starts or do you mean something else?

I just want to make it entirely dynamic and automatic. Looking at https://hieromon.github.io/AutoConnect/index.html but there are more lightweight options...
The web interface shouldn't really care how it's on-line (connected to a configured WiFi or in Captive portal mode on it's own AP) and when one of the configured WiFi networks come online it should just auto-connect there.

Right now it blocks while it tries and after timeout it goes offline

tbaumann:
> I think mDNS might not be enabled automatically with ESP32? With ESP8266 it already works fine without explicitly adding it. You mean at boot it would be nice if it started heating right away while also trying to connect to WiFi and retrying as well if not successful in parallel? Would be nice but also might be a bit tricky to do things in parallel?

The A record is automatic the SRV record not. That's why I added it

kjyv:
Aha, I see.
Anyway, the AP logic is intended to let you reconfigure you WiFi config in case it can't connect. The workflow you describe is not straight forward to use as it would mean you still see the normal website in AP mode (which could be a useful to have optional feature though in case there is no wifi to use the machine with) - how would you change wifi credentials in that case using the WifiManager UI? I guess we could have a reset button in the website somewhere

tbaumann:
> Aha, I see. Anyway, the AP logic is intended to let you reconfigure you WiFi config in case it can't connect. The workflow you describe is not straight forward to use as it would mean you still see the normal website in AP mode (which could be a useful to have optional feature though in case there is no wifi to use the machine with) - how would you change wifi credentials in that case using the WifiManager UI? I guess we could have a reset button in the website somewhere

Perhaps not the best place to discuss here. But of course there would be a settings page to select the wifi.

And even if we don't set it up like this. There is absolutely no need to block the boot process. The wifi manager can give you a callback on fail and success.

kjyv:
Not the best place since it is not related to the temp sensor true, but just one last remark: The blocking is not ideal true, but apart from not being connected, it should mainly not keep the machine from heating until the portal times out. The fact that it does not connect after a few retries and the set timeout is probably not improved by retrying to connect again and again. Ideally it should just connect every time.

tbaumann:
> Not the best place since it is not related to the temp sensor true, but just one last remark: The blocking is not ideal true, but apart from not being connected, it should mainly not keep the machine from heating until the portal times out. The fact that it does not connect after a few retries and the set timeout is probably not improved by retrying to connect again and again. Ideally it should just connect every time.

I might be wrong, but I don't think it's running before it connects. At least that is when the debug output starts. (I don't have my display yet and only a bench setup.)

kjyv:
I might have put that a bit weird. I meant that the code currently does not start heating before it is connected, but it ideally should because it delays the startup a lot if for some reason the connection is not possible. I think that is the main thing to fix in terms of WifiManager, otherwise I don't see a problem with the current logic.

tbaumann:
> I try the PR and get a build error. Can you check your PR again, please? `src/rancilio-pid.cpp:788:1: error: expected declaration before '}' token` I find the wrong } and correct it, can you please check if everything works with the our sensor?

I will rebase it without mdns support (and without the syntax error you added :D )

tbaumann:
Run it again. Works for me

tbaumann:
Ah you found the bracket. :D 
I will try to give it a test tomorrow. But I'm quite sure it will be fine. That one merged commit is totally harmless.

murmeltier08:
I have the problem that my machine stucks while booting now.

LoQue90:
@tbaumann would you like to look over your PR?
rebase and stuff? so we could merge this
wich sensor do I need to test this?
Thanks!

tbaumann:
Sorry for the late replay. Busy times.

I no longer use rancilio-pid on my machine. I don't have a need for this right now. If you merge it, you have to maintain it.
But it might be worth it. It's a much more reliable option to thermocouples. Physically and in terms of electrical interference.

I'm happy to rebase it for you. Give me a few days...

fiendie:
Closing because PR is stale and doesn't work with the current pin mapping on our PCB.

PR #272, opened 2022-10-19
Quick Mill Cold Start Kp

cottec:
To fix Cold Start Kp being too low for Quick Mill Machines


kjyv:
The max value should maybe be a bit larger that the value from the manual - How about 350?

cottec:
feel free, shouldnt harm anything, but not sure if more is needed at all :)

kjyv:
I can't change your pull request ;)

cottec:
ups, changed :)

PR #277, opened 2022-11-04
Remove ESP8266 support, clean up repository

fiendie:
* Remove ESP822 support
* Move hardware-related stuff to hardware repository
* Move (most of) InfluxDB and MQTT code out of the main sketch
* Use #pragma once instead  of include guards in all header files
* Introduce default pin mapping, remove pin definitions from userConfig.h
* Remove ToF Sensor support in preparation for water level sensor
* Remove E-Trigger support in preparation for standby mode


fiendie:
Moved PR to https://github.com/rancilio-pid/clevercoffee-esp32

PR #295, opened 2022-12-28
Use MQTT will for availability

Nilaik:
* change the will's topic name from will to available
* send offline as last will
* when sending all system parameters also send online for the will topic

### DIFF
--- a/src/rancilio-pid.cpp
+++ b/src/rancilio-pid.cpp
@@ -2319,7 +2319,7 @@ void writeSysParamsToMQTT(void) {
 
         if (mqtt.connected() == 1) {
             // status topic (will sets it to offline)
-            mqtt_publish("status", "online");
+            mqtt_publish("status", (char *)"online");
 
             mqtt_publish("temperature", number2string(temperature));
             mqtt_publish("brewSetPoint", number2string(brewSetPoint));
--- a/src/rancilio-pid.cpp
+++ b/src/rancilio-pid.cpp
@@ -1832,7 +1832,7 @@ void setup() {
         }
 
         if (MQTT == 1) {
-            snprintf(topic_will, sizeof(topic_will), "%s%s/%s", mqtt_topic_prefix, hostname, "available");
+            snprintf(topic_will, sizeof(topic_will), "%s%s/%s", mqtt_topic_prefix, hostname, "status");
             snprintf(topic_set, sizeof(topic_set), "%s%s/+/%s", mqtt_topic_prefix, hostname, "set");
             mqtt.setServer(mqtt_server_ip, mqtt_server_port);
             mqtt.setCallback(mqtt_callback);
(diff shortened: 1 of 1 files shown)


Nilaik:
The last will was already implemented for the topic will.

I created a pull request where I changed the will topic from "will" to "available". Also instead of sending "exit" I send "offline" as payload. In order to indicate the online status the payload "online" is send in the function writeSysParamsToMQTT.

Not sure if this is an allowed change since this would be a breaking change for version 3.1.

src/rancilio-pid.cpp
```diff
@@ -1832,7 +1832,7 @@ void setup() {
         }
 
         if (MQTT == 1) {
-            snprintf(topic_will, sizeof(topic_will), "%s%s/%s", mqtt_topic_prefix, hostname, "will");
+            snprintf(topic_will, sizeof(topic_will), "%s%s/%s", mqtt_topic_prefix, hostname, "available");
```
LoQue90, line 1835: snprintf(topic_will, sizeof(topic_will), "%s%s/%s", mqtt_topic_prefix, hostname, "status");
I would change the topic name.
Compared to other mqtt stuff that i use, following would make more sense to me:
Topic will be "status", payload will be "online" or "offline"
Logic in my mind: If i want to get the availability i check the status of the mashine online or offline.
If the status of the mashine is online it is available.

src/rancilio-pid.cpp
```diff
@@ -2318,6 +2318,9 @@ void writeSysParamsToMQTT(void) {
         previousMillisMQTT = currentMillisMQTT;
 
         if (mqtt.connected() == 1) {
+            // available topic (will sets it to offline)
```
LoQue90, line 2321: // status topic (will sets it to offline)

src/rancilio-pid.cpp
```diff
@@ -2318,6 +2318,9 @@ void writeSysParamsToMQTT(void) {
...
+            mqtt_publish("available", "online");
```
LoQue90, line 2322: mqtt_publish("status", "online");

LoQue90 (changes requested):
Works fine for me, would like to see a merge to master with my suggestions.
Just have to see on Home Assistant side how to get a nice readout for the temperature when stauts=offline

src/rancilio-pid.cpp
Nilaik: Sounds good to me. I changed the topic name of the will to status and tested it with my setup.

Nilaik: In Home Assistant I have defined my temperature sensor as following:

```
  - name: Rancilio Silvia Temperature
    unique_id: sensor.silvia_temp
    availability:
      - payload_available: online
        payload_not_available: offline
        topic: espresso/rancilio.silvia/status
    state_topic: espresso/rancilio.silvia/temperature
    state_class: measurement
    device_class: temperature
    unit_of_measurement: °C
```

After switching my silvia off home assistant receives the will message and marks the sensor as unavailable.

LoQue90 (approved):
I would say fine fore merge.
Thank you!

PR #298, opened 2023-01-02
Add ToF sensor data/water level to MQTT

Nilaik:
Water level was not published via MQTT.

I also re-positioned the water level percentage on the display. It was 1 pixel lower than the MQTT string and touched the MQTT string horizontally. So I moved it up by 1 pixel and to the right by 5 pixel.

Before:

image

After:

image


LoQue90:
looks fine!

LoQue90 (approved):

PR #300, opened 2023-01-04
Add machine state to mqtt

Nilaik:


LoQue90 (approved):
runs fine, but some states don´t get published because there are no up long enough

Nilaik:
Yes that is true. In the linked issue I mentioned an alternative but am not sure if this is the right place to do it:

> Sending the machine state could also be added to printMachineState but that felt kind of weird since this function only contains debug prints. We would also need to duplicate the MQTT check in printMachineState so it is only called when MQTT is configured and connected.

LoQue90:
Taking it from the debug print sounds not like a good idea. 
Until now I don't know what the mashine states in mqtt are good for. But the pid states are all up long enough to publish. 

> Yes that is true. In the linked issue I mentioned an alternative but am not sure if this is the right place to do it:
> 
> > Sending the machine state could also be added to printMachineState but that felt kind of weird since this function only contains debug prints. We would also need to duplicate the MQTT check in printMachineState so it is only called when MQTT is configured and connected.

Nilaik:
I think it might be useful for automation in Home Assistant. For example I have a lamp at my desk that starts flashing when the target temperature is reached + 15 minutes waiting time, because I regularly start the machine continue working and forget to make my Espresso 🙈

At the moment I created my own helper state on the home assistant site when the target temperature is reached. But it might be nice to switch to the actual state the machine can publish like kPidNormal. 

Also Emergency Stop and Sensor Error are states that would be interesting for me to have in Home Assistant.

PR #303, opened 2023-01-10
Added Calibration via MQTT and Display

maunsen:
I added weights to MQTT and added an option to calibrate the scale via setting the scaleCalibration Value to 1 via mqtt.
First you need to weigh a known weight and add its weight to the scaleKnownWeigth Value via MQTT.
The Display guides you through the calbration and tells you when to add your known weight.
It then does the calibration and adds the new calibrationValue to storage.


src/brewscaleini.h
```diff
@@ -28,14 +28,16 @@ boolean brewPIDdisabled = false;                    // is PID disabled for delay
...
 // Shot timer with or without scale
 #if (ONLYPIDSCALE == 1 || BREWMODE == 2)
     int shottimercounter = 10 ;
-    float calibrationValue = SCALE_CALIBRATION_FACTOR;  // use calibration example to get value
+    
     float weight = 0;                                   // value from HX711
     float weightPreBrew = 0;                            // value of scale before wrew started
     float weightBrew = 0;                               // weight value of brew
-    float scaleDelayValue = 2.5;                        // value in gramm that takes still flows onto the scale after brew is stopped
+    float scaleDelayValue = 0.5;                        // value in gramm that takes still flows onto the scale after brew is stopped
```
kjyv, line 40:
```suggestion
    float scaleDelayValue = 0.5;                        // value in grams that takes still flows onto the scale after brew is stopped
```

src/Storage.h
```diff
@@ -39,8 +39,11 @@ typedef enum
   STO_ITEM_BREW_TEMP_OFFSET,        // brew temp offset
   STO_ITEM_PID_START_PONM,          // Use PonM for cold start phase (otherwise use normal PID and same params)
   STO_ITEM_USE_BD_PID,              // use separate PID for brew detection (otherwise continue with regular PID)
+  STO_ITEM_SCALECALIBRATION,         // Calibration Value for Scale
+  STO_ITEM_SCALEKNOWNWEIGHT,         // Known Weight for Scale Calibration
```
kjyv, line 43:
```suggestion
  STO_ITEM_SCALEKNOWNWEIGHT,        // Known weight for scale calibration
```

src/Storage.h
```diff
@@ -39,8 +39,11 @@ typedef enum
   STO_ITEM_BREW_TEMP_OFFSET,        // brew temp offset
   STO_ITEM_PID_START_PONM,          // Use PonM for cold start phase (otherwise use normal PID and same params)
   STO_ITEM_USE_BD_PID,              // use separate PID for brew detection (otherwise continue with regular PID)
+  STO_ITEM_SCALECALIBRATION,         // Calibration Value for Scale
```
kjyv, line 42:
```suggestion
  STO_ITEM_SCALECALIBRATION,         // Calibration Value for Scale
```
```suggestion
  STO_ITEM_SCALECALIBRATION,        // Calibration value for scale
```

src/brewscaleini.h
```diff
@@ -28,14 +28,16 @@ boolean brewPIDdisabled = false;                    // is PID disabled for delay
 const unsigned long analogreadingtimeinterval = 10; // ms
 unsigned long previousMillistempanalogreading;      // ms for analogreading
 
+
+
```
kjyv, line 32: don't add random empty lines

src/brewvoid.h
```diff
@@ -366,6 +366,7 @@ void brew() {
 
         // state machine for brew
         switch (brewcounter) {
+
```
kjyv, line 369: also here, please no added empty lines

src/rancilio-pid.cpp
```diff
@@ -989,6 +990,10 @@ void assignMQTTParam(char *param, double value) {
                         *(uint8_t *)m.mqttVarPtr = value;
                         paramValid = true;
                         break;
+                    case tFloat:
```
kjyv, line 993: why add extra float for this one variable? just make it double as well?

src/rancilio-pid.cpp
```diff
@@ -1920,6 +1932,11 @@ void setup() {
         previousMillisPressure = currentTime;
     #endif
 
+            // Init Scale by BREWMODE 2 or SHOTTIMER 2
```
kjyv, line 1935:
```suggestion
    // Init Scale by BREWMODE 2 or SHOTTIMER 2
```

kjyv: comment is a bit redundant but also doesn't match the condition on the next line

src/rancilio-pid.cpp
```diff
@@ -2022,6 +2039,7 @@ void looppid() {
     if ((millis() - lastTempEvent) > tempEventInterval) {
         //send temperatures to website endpoint
         sendTempEvent(temperature, brewSetPoint, pidOutput);
+        sendWeightEvent(weight, weightBrew, 22.22);
```
kjyv, line 2042: why is it sending the fixed value 22.22?

src/rancilio-pid.h
```diff
@@ -33,6 +33,8 @@ int writeSysParamsToStorage(void);
 #define PRE_INFUSION_TIME 2        // pre-infusion time in seconds
 #define PRE_INFUSION_PAUSE_TIME 5  // pre-infusion pause time in seconds
 #define SCALE_WEIGHTSETPOINT 30    // Target weight in grams
+#define SCALE_CALIBRATION -250.00  // Calibration Value
```
kjyv, line 36:
```suggestion
#define SCALE_CALIBRATION -250.00  // Calibration value
```

src/rancilio-pid.h
```diff
@@ -33,6 +33,8 @@ int writeSysParamsToStorage(void);
...
+#define SCALE_KNOWNWEIGHT 200      // Known Weight in grams for calibrating the scale
```
kjyv, line 37:
```suggestion
#define SCALE_KNOWNWEIGHT 200      // Known weight in grams for calibrating the scale
```

src/scalevoid.h
```diff
@@ -36,17 +36,28 @@ void checkWeight() {
  * @brief Initialize scale
  */
 void initScale() {
-    LoadCell.begin();
-    long stabilizingtime = 2000; // tare preciscion can be improved by adding a few seconds of stabilizing time
+    String scaleCalibrationString = number2string(scaleCalibration);
+    String StringCompare = "1.00";
```
kjyv, line 40:
```suggestion
    String stringCompare = "1.00";
```

src/scalevoid.h
```diff
@@ -36,17 +36,28 @@ void checkWeight() {
...
+    boolean calibration = false;
+    LoadCell.begin(128);
+    long stabilizingtime = 4000; // tare preciscion can be improved by adding a few seconds of stabilizing time
```
kjyv, line 43:
```suggestion
    int stabilizationTime = 4000; // tare precision can be improved by adding a few seconds of stabilization time
```

src/scalevoid.h
```diff
@@ -36,17 +36,28 @@ void checkWeight() {
...
     boolean _tare = true; //set this to false if you don't want tare to be performed in the next step
 
     u8g2.clearBuffer();
     u8g2.drawStr(0, 2, "Taring scale,");
     u8g2.drawStr(0, 12, "remove any load!");
-    u8g2.drawStr(0, 22, "....");
-    delay(2000);
+    
+    if(scaleCalibrationString == StringCompare){
```
kjyv, line 50: why is this comparison done with strings? should be numbers if possible to save memory

src/scalevoid.h
```diff
@@ -56,11 +67,34 @@ void initScale() {
         u8g2.sendBuffer();
     }
     else {
-        u8g2.drawStr(0, 32, "done.");
-        u8g2.sendBuffer();
+        if(calibration){
```
kjyv, line 70:
```suggestion
        if (calibration){
```

src/rancilio-pid.cpp
```diff
@@ -659,7 +615,52 @@ void refreshTemp() {
         }
     }
 }
+boolean sys = readSysParamsFromStorage();
```
kjyv, line 618: why move all of this down?

src/Storage.h
kjyv: STO_ITEM_SCALE_KNOWN_WEIGHT would be a better name

data/js/temp.js
```diff
@@ -301,6 +301,15 @@ if (!!window.EventSource) {
 
             document.getElementById("varTEMP").innerText = myObj["currentTemp"].toFixed(1)
         },
+        'new_weights',
```
kjyv, line 304: Does this work? Should probably be in a separate source.addEventListener() call

src/scalevoid.h
maunsen: Problem was, that you cant compare floats. Better solution would be to transform to double.

src/rancilio-pid.cpp
maunsen: that is my fault, that was just for testing. Should be weightPreBrew

src/rancilio-pid.cpp
maunsen: good idea. I will rework the code to make it double. Solves the problem in the if statement later on

src/rancilio-pid.cpp
maunsen: To make sure that the SysParams are loaded before the scale is initiated.

.gitignore
```diff
@@ -12,3 +12,4 @@ src/userConfig.h
 
 platformio_extra.ini
 **/rancilio-pid.ino.cpp
+src/userConfig_sample.h
```
kjyv, line 15:
```suggestion
```

maunsen: Done

src/scalevoid.h
maunsen: can't use double als the calibration Value needs to be float.
So saved value as float but cast to double in if statement

kjyv: You can compare floats of course. Doubles are unnecessary here because you don't need that high accuracy

kjyv: if you need to check for equality you usually do something like this:
if(fabs(float1 - float2) < 0.00001)

maunsen: Is this solution acceptable? 
if ((double) scaleCalibration == 1.00)

kjyv: no, it's not guaranteed that scaleCalibration is exactly the double number 1.00. It could be 0.999999. Only if there is code that sets it to 1.00 you can assume that. 
You always need to compare the difference with a small epsilon like above otherwise

maunsen: It is set to 1.00 via mqtt in the process.
I am using this code for a few days and it is working flawless.

When you set it to 1.00 via mqtt the calibration process is started and an this process it is set to the correct value.
In the hx711-lib it is also advised to use 1.00 as the inital value to start the calibration.

Shall I change it anyway? Then we would need another float variable which ist set to 1.00 for the if-statement.

kjyv: just use `if (fabs(scaleCalibration - 1) < 0.00001) {`

maunsen: Oh, thats nice! Thanks!

kjyv: or abs if you don't want to use a float variable - it's a bit weird that our code uses double for almost everything for no reason

src/rancilio-pid.cpp
maunsen: It needed to be float as the hx711 lib needs a float as calibration value

kjyv: you can just cast a float to double...

maunsen: But can I cast a double back to float?

kjyv: You can cast in both directions as you like, yes. From double to float, precision loss will occur but doesn't matter here

murmeltier08:
Rebase to 4.0.0 is necessary

fiendie:
@maunsen: In its current state I don't really see a path forward. There are now 74 in your branch, including a lot of duplicates intertwined with commits from the master branch. I don't exactly know what went wrong but at this point the easiest solution might be to back up your changes and applying them to the latest master by hand.

fiendie:
Will be re-evaluated in a new PR.

PR #304, opened 2023-01-19
fix crash while setup wifi via wifi manager

murmeltier08:
Wifi manager crash with ESP8266 because of the further display initialization.
PR should solve it. Display initialization in the end of the setup().

### DIFF
--- a/src/rancilio-pid.cpp
+++ b/src/rancilio-pid.cpp
@@ -1611,9 +1611,9 @@ void wiFiSetup() {
     } else {
         debugPrintln("WiFi connection timed out...");
 
-        #if OLED_DISPLAY != 0
-            displayLogo(langstring_nowifi[0], langstring_nowifi[1]);
-        #endif
+       // #if OLED_DISPLAY != 0
+       //     displayLogo(langstring_nowifi[0], langstring_nowifi[1]);
+       // #endif
 
         wm.disconnect();
         delay(1000);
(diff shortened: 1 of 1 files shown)


src/rancilio-pid.cpp
```diff
@@ -1813,15 +1813,7 @@ void setup() {
...
-    #if OLED_DISPLAY != 0
-        u8g2.setI2CAddress(oled_i2c * 2);
-        u8g2.begin();
-        u8g2_prepare();
-        displayLogo(String("Version ") + String(sysVersion), "");
-       // delay(2000); // caused crash with wifi manager
-    #endif
-
-    // Init Scale by BREWMODE 2 or SHOTTIMER 2
+        // Init Scale by BREWMODE 2 or SHOTTIMER 2
```
kjyv, line 1816:
```suggestion
    // Init Scale by BREWMODE 2 or SHOTTIMER 2
```

src/rancilio-pid.cpp
```diff
@@ -1923,6 +1915,16 @@ void setup() {
     setupDone = true;
 
     enableTimer1();
+    // Display has to be init in the end, because ESP8266 crash during wifi manager setup.
```
kjyv, line 1918:
```suggestion
    // Display has to be initialized after the display because ESP8266 can crash during wifi manager setup. 
```

murmeltier08:
new PR for ESP8266

PR #313, opened 2023-01-24
Remove obsolete TSIC library

fiendie:


fiendie:
@LoQue90 was faster 😄

PR #318, opened 2023-01-25
Replace mqttVars vector with map

Alcoinus:
This isn't tested, as I use neither software BD nor scale. But as the changes are small, I believe there won't be any issues.

EDIT:
I continued to work on this and replaced the mqttVars vector with a map. This is intended to be a precursor to allow Home Assistant MQTT auto discovery.

This time, I have tested the changes and they work without issue for me ;)

### DIFF
--- a/src/debugSerial.cpp
+++ b/src/debugSerial.cpp
@@ -52,6 +52,11 @@ void debugPrintln(const char *message) {
     }
 }
 
+// Print to remote serial (e.g. using OTA Monitor Task ) if client is connected, otherwise use hardware serial
+void debugPrintln(const String& message) {
+    debugPrintln(message.c_str());
+}
+
 void debugPrint(const char *message) {
     char time[12];
     getCurrentTimeString(time);
@@ -62,6 +67,10 @@ void debugPrint(const char *message) {
         Serial.print(time);
         Serial.print(message);
     }
+}
+
+void debugPrint(const String& message) {
+    debugPrint(message.c_str());
 }
 
 size_t debugPrintf(const char *format, ...) {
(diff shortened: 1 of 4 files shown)


src/debugSerial.cpp
```diff
@@ -52,6 +52,19 @@ void debugPrintln(const char *message) {
     }
 }
 
+// Print to remote serial (e.g. using OTA Monitor Task ) if client is connected, otherwise use hardware serial
+void debugPrintln(const String& message) {
+    char time[12];
```
kjyv, line 57: better to not repeat basically the whole content of debugPrintln(const char *message) here, just call it:
debugPrintln(message.c_str());

Alcoinus: Thanks, I have updated it.

kjyv (approved):

src/main.cpp
```diff
@@ -1818,6 +1808,46 @@ void setup() {
...
+    mqttVars["backflushON"] = []{ return &editableVars.at("BACKFLUSH_ON"); };
+    mqttVars["startUsePonM"] = []{ return &editableVars.at("START_USE_PONM"); };
+    mqttVars["startKp"] = []{ return &editableVars.at("START_KP"); };
+    mqttVars["startTn"] = []{ return &editableVars.at("START_TN"); };
+    mqttVars["aggKp"] = []{ return &editableVars.at("PID_KP"); };
+    mqttVars["aggTn"] = []{ return &editableVars.at("PID_TN"); };
+    mqttVars["aggTv"] = []{ return &editableVars.at("PID_TV"); };
+    mqttVars["aggIMax"] = []{ return &editableVars.at("PID_I_MAX"); };
+    mqttVars["steamKp"] = []{ return &editableVars.at("STEAM_KP"); };
+    if (ONLYPID == 0) {
```
fiendie, line 1827: Could use some blank lines here and there :)
We usually have one before an if-statement.

Alcoinus: Thanks for the input. I have added some blank lines.

fiendie (dismissed):

PR #320, opened 2023-01-29
Fixed scale pin names

Invenis:
Fixed scale pin references after naming convention changed alongside other ESP32 changes.

### DIFF
--- a/src/brewscaleini.h
+++ b/src/brewscaleini.h
@@ -6,6 +6,7 @@
  */
 
 #pragma once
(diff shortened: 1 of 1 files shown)


fiendie:
We changed the pinmapping again yesterday. If it's OK with you I would just incorporate your commit into that change.

Invenis:
Sure, no problem

Invenis:
Merged as a part of #321.

PR #321, opened 2023-01-29
Fix pin definitions for scale, update pin mapping to match latest revision of the PCB

fiendie:
@Invenis: I cherry-picked your commit before making the other changes. Could you close PR #320? Thanks!


Invenis:
Will do, thanks!

fiendie:
> Will do, thanks!

Thanks for the find! :)

kjyv (approved):

PR #326, opened 2023-02-04
Allow using webinterface without internet

kjyv:
* load js libraries from filesystem instead of internet by default (change NOINTERNET in code for previous behaviour)
* use uPlot instead of plotly (48kb instead of ~950kb) so everything actually fits on the filesystem
* add clevercoffee logo and all necessary js, css and woff files


src/userConfig_sample.h
```diff
@@ -49,6 +49,7 @@ enum MACHINE {
 #define PASS "CleverCoffee"        // default password for WiFiManager
 #define MAXWIFIRECONNECTS 5        // maximum number of reconnection attempts, use -1 to deactivate
 #define WIFICONNECTIONDELAY 10000  // delay between reconnects in ms
+#define NOINTERNET 1               // Use libraries for the webinterface from the internet (0) or from the local filesystem (1). 0 has slightly faster load times
```
murmeltier08, line 52: We have CONNECTMODE = 0 (offline). What is the different to NOINTERNET?

kjyv: The difference is that CONNECTMODE == 0 does not attempt to connect to Wifi while this is just not loading any dependencies for the website from the internet. Maybe we can name them differently to distinguish better

fiendie (approved):

src/userConfig_sample.h
fiendie: I think we should remove this line and make it the default as we discussed.

kjyv: If we're sure that most people won't need the previous behaviour then sure. We can just allow changing it in the code for those rare cases

fiendie (dismissed):

murmeltier08:
I get following message. I do erase flash, load the new filesystem, try to upload the firmware:
Building in release mode
Retrieving maximum program size .pio/build/esp32_usb/firmware.elf
Checking size .pio/build/esp32_usb/firmware.elf
Advanced Memory Usage is available via "PlatformIO Home > Project Inspect"
Error: The program size (1258937 bytes) is greater than maximum allowed (1245184 bytes)
*** [checkprogsize] Explicit exit, status 1
RAM:   [=         ]  10.8% (used 57624 bytes from 532480 bytes)
Flash: [==========]  101.1% (used 1258937 bytes from 1245184 bytes)

kjyv:
Do you still have something else in the filesystem folder? I only see about 850kb that are used.

kjyv:
Please try again, should be ok also with influx now

murmeltier08:
Works now fine.

PR #331, opened 2023-02-08
Use larger partition scheme, add check_tool, move identical entries to global env

fiendie:


platformio.ini
```diff
@@ -4,6 +4,14 @@ src_dir = src
 extra_configs = platformio_extra.ini
 
 [env]
+platform = espressif32 @^5
+board = az-delivery-devkit-v4
+board_build.filesystem = littlefs
+board_build.partitions = huge_app.csv
```
kjyv, line 10: I think huge_app doesn't allow ota anymore

kjyv: Compared to https://github.com/espressif/arduino-esp32/blob/master/tools/partitions/no_ota.csv, the code partition is a bit smaller and spiffs is larger

kjyv: I've more or less maximized the filesystem in PR #326 already

fiendie: default.csv then, partitions_4M.csv is a partition scheme part of Espressif IDF not the Arduino framework.

kjyv: It's a file that's in the PR

fiendie: Either way, it works fine with default.csv, on PR #326 as well.

PR #333, opened 2023-02-09
Scale configuration

maunsen:
Added parameters to be able to calibrate the scale via mqtt an webinterface.
Also added options to tare and calibrate the scale from the webinterface


fiendie:
Commits from this pull request have been added to #421

PR #335, opened 2023-02-15
adding trigger for Steam mode

LoQue90:
Steamswitch has now 3 modes:
off, switch or trigger
removed double pinmode declaration for Steamswitch


LoQue90:
now part of pr#342

PR #339, opened 2023-02-16
Add machine state to mqtt again

Alcoinus:


src/MQTT.h
```diff
@@ -76,10 +76,8 @@ bool mqtt_publish(const char *reading, char *payload, boolean retain) {
 /**
  * @brief Publish Data to MQTT
  */
-bool mqtt_publish(const char *reading, char *payload) {
-    char topic[120];
-    snprintf(topic, 120, "%s%s/%s", mqtt_topic_prefix, hostname, reading);
-    return mqtt.publish(topic, payload, false);
+bool mqtt_publish(const char *reading, char const *payload) {
+    return mqtt_publish(reading, payload, false);
```
kjyv, line 80: This method doesn't do anything anymore, so can be removed. But is it really correct to remove the topic etc.?

src/MQTT.h
```diff
@@ -192,7 +190,7 @@ void writeSysParamsToMQTT(void) {
             }
             
             for (const auto& pair : mqttSensors) {
-                mqtt_publish(pair.first, number2string(pair.second()));
+                mqtt_publish(pair.first, pair.second());
```
kjyv, line 193: Why this change? it's shorter to just convert the numbers here rather than for each mqttSensors entry

src/MQTT.h
kjyv: Or even just add an overload for mqtt_publish that accepts doubles

src/MQTT.h
kjyv: Or even just add an overload for mqtt_publish that accepts doubles

src/MQTT.h
Alcoinus: In 03043fbe4efc34a664eff47dddc81b53f7b502b8 I overloaded this function to allow for explicit setting of the retain behaviour. At the time, I just duplicated the code, but afterwards realized, that it could be unified this way.

(just to be sure: you have noticed, that it is mqtt_publish (underscore) now instead of mqtt.publish (dot)?)

src/MQTT.h
Alcoinus: > Or even just add an overload for mqtt_publish that accepts doubles

That would be nice. But this would require a modification of the value type of the `mqttSensors` map. The function has one specific return type, but next to the `double` values, `machineStateString` topic needs to return a `char *`. At the moment I do not have another solution for it.

src/MQTT.h
kjyv: Ah ok, I did overlook that it changed from . to _. So topic is fine, but the method is still not really doing anything except hiding the retain parameter.

kjyv: You can also define the same by
`bool mqtt_publish(const char *reading, char const *payload, boolean retain = false) {
`

src/MQTT.h
kjyv: I'm not sure I understand what you mean. If you add an overload of mqtt_publish that accepts a double (that can then call number2string), you can call it with pair.second() here without number2string and just leave the type of mqttSensors as double as it was. This would reduce unnecessary memory usage of mqttSensors for all the strings.

src/MQTT.h
Alcoinus: Ok thank you for the advice, I have modified it accordingly.

src/MQTT.h
Alcoinus: The problem is line 330 of `main.cpp`:
`std::map<const char*, std::function<char const*()>, cmp_str> mqttSensors = {};`

The map is defined so that pair.second is a function, which returns a `char const*`. I would need to find a way, to let pair.second return either a `double` or a `char const*` or in the future any other datatype, which might additionally be included in the map.

I do not have another solution for it at the moment, but I am happy for any suggestion.

LoQue90:
@Alcoinus merging hassio auto discovery, mqtt changed a lot.
maybe you want to start from there with adding the mashine state back to mqtt?

Alcoinus:
I'm gonna take a look at it in the next couple of days.

Alcoinus:
Closing, as the changes are now in #385

PR #340, opened 2023-02-19
DRAFT / TEST Wifi flag for display to prevent crash

murmeltier08:
Draft for testing of the wifi flag. But not show a improvement.


LoQue90:
isn´t this stale and could be closed, because we made the decision to remove WM on total for esp8266?

kjyv:
It seems also this is party already on master?
e.g. STO_ITEM_WIFI_CREDENTIALS_SAVED is existent there

LoQue90:
> It seems also this is party already on master? e.g. STO_ITEM_WIFI_CREDENTIALS_SAVED is existent there

true, should be taken out :)

Issue #347, opened 2023-04-16
brew by weight and no display not possible


LoQue90:
when brewmode = 2 and display = 0, build failed because void initScale() still try to output on display

LoQue90:
fixed in #348

PR #348, opened 2023-04-22
bugfix for #347

murmeltier08:
Only bug fixing for Issue #347 , we have to solve it better in the ESP32 master. We use now pre processor commands.

### DIFF
--- a/src/scalevoid.h
+++ b/src/scalevoid.h
@@ -60,7 +60,7 @@ void initScale() {
         #endif
         
     } else {
-         #if (OLED_DISPLAY != 0)
+        #if (OLED_DISPLAY != 0)
         u8g2.drawStr(0, 32, "done.");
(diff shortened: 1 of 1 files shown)


src/scalevoid.h
```diff
@@ -40,24 +40,30 @@ void initScale() {
...
-    else {
+         #if (OLED_DISPLAY != 0)
+            u8g2.drawStr(0, 32, "failed!");
+            u8g2.drawStr(0, 42, "Scale not working...");    // scale timeout will most likely trigger after OTA update, but will still work after boot
+            delay(5000);
+            u8g2.sendBuffer();
+        #endif
+        
+    } else {
+         #if (OLED_DISPLAY != 0)
```
kjyv, line 63:
```suggestion
        #if (OLED_DISPLAY != 0)
```

PR #350, opened 2023-05-03
pinmapping based on MCU type

LoQue90:
moved pinmapping out of userconfig, pinmapping based on MCU TYPE automatically
for esp8266 board manual changes based on feature set are still necessary, has to be done in "esp8266nodemcuv2.h"
change standard PINBREWSWITCH for esp8266, because PIN A0 mostly causes problems, PIN 15 is the better choise

### DIFF
--- a/src/ISR.cpp
+++ b/src/ISR.cpp
@@ -1,6 +1,6 @@
 #include <Arduino.h>
 #include "ISR.h"
-#include "userConfig.h"
+#include "hardware.h"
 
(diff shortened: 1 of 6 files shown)


src/ISR.cpp
```diff
@@ -1,6 +1,7 @@
 #include <Arduino.h>
 #include "ISR.h"
 #include "userConfig.h"
```
fiendie, line 3: This include can be removed now

LoQue90: done, thanks!

PR #357, opened 2023-05-29
Added manual Brewmode

dafes:
* Added Brewmode 3 to have "full" control over the Brewing (Switch controls Brew Start/Stop)

### DIFF
--- a/src/brewvoid.h
+++ b/src/brewvoid.h
@@ -429,3 +429,118 @@ void brew() {
     }
 }
 #endif
+
+
+#if (BREWMODE == 3)  // manual brew mode based on switch
+/**
+ * @brief PreInfusion, Brew manually
+ */
+void brew() {
+    if (OnlyPID == 0) {
+        unsigned long currentMillistemp = millis();
+        checkbrewswitch();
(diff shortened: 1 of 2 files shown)


fiendie:
I see what you are getting at. But wouldn't that essentially be the same as just setting the pre-infusion pause to 0 and the brew time to something longer than you need? You can already stop the brew at any time.

LoQue90:
there is no need to build/upload to change brewtime or pause. You can do it live on the website.
i think a manual brew mode is not realy necessary, because you can achive the same with the way from fiendie.
But it´s not a big deal to implement it. 
What i would like to have then:
brewtimer should be running while brewing and should have the same behaviour like "brew by time" without auto stop
switch and trigger should be implemented
Gonna look at what you have done :)

fiendie:
I would love a more straightforward way do disable pre-infusion/timed brewing as well. We wanted to start moving stuff out of the userConfig ever since ditching ESP8266 support so it might be worth refactoring the whole brew mode code. 
You should come over to the Mattermost chat so we can compare notes :)

PR #358, opened 2023-05-29
Added InfluxDB retries 

dafes:
* added InfluxDB Check to disable InfluxDB inflight before blocking the ESP


fiendie (approved):
Had it running for a few days now without any issues.

PR #363, opened 2023-07-20
Fix because bootstrap JS wasnt loading

SphaeroX:
integrity hash was wrong, bs not loaded. removed for local. 

I am in favor of always loading the files locally from ESP32 without online verification.


murmeltier08:
did someone tested it?

PR #371, opened 2023-09-08
fix: tempature sensor configuration for DS18B20

cellcortex:
fixing up the config for the DS18B20 temperature sensor. 
* The include of the Dallas drivers was before the config import - so it wouldn't pickup the precompiler switch
* There was a typo in the PIN constant name

I also removed a bit of code duplication in the `refreshTemp`

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -12,6 +12,9 @@
 #define FW_HOTFIX     0
 #define FW_BRANCH     "MASTER"
 
+// config (include early so that precompiler flags are set)
+#include "userConfig.h"             // needs to be configured by the user
+
 // Libraries
 #include <ArduinoOTA.h>
 #include <map>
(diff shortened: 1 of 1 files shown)


LoQue90:
anybody who using a DS18B20 sensor is realy welcome to test this PR

src/main.cpp
```diff
@@ -92,7 +94,6 @@ int connectmode = CONNECTMODE;
 
 int offlineMode = 0;
 const int OnlyPID = ONLYPID;
-const int TempSensor = TEMPSENSOR;
```
LoQue90, line 95: as far as I know we need the TEMPSENSOR constant macro as a int in the code, so we can change it thru the webseite later

murmeltier08:
Also a test with a normal TSIC will be necessary.

LoQue90:
we gona add a HAL which makes sensor support much easier.
Lets fix ds18B20 sensor after this

LoQue90:
#458 will handleds18B20 support, lets check your fixes after adding the HAL

LoQue90:
should be fix with #458

PR #382, opened 2023-10-14
WIP: Add settings menu with rotary encoder

FabianSperrle:
- define menu structure for setting brew temp and steam temp (other entries for times etc. could easily be added)
- add rotary encoder library to read encoder state

TODO: 
- only works with ESP32 for now due to the timers needed for the encoder --> Should probably be ported to master? I started the implementation on the wrong branch / the state that I use in my machine
- no support for upright displays yet
- multilanguage support started but not yet finished
- Pins need to be defined in `userConfig.h` at the moment

I don't like that all the code is in the "main" file, making it even longer and more cluttered. However, I did not manage to move the display rendering code to `displays.h` yet as I always faced issues with variables not being defined. 

I'm curious to hear your feedback and ideas on how to proceed with this, if you are interested.


FabianSperrle:
Ported to master, closing here.

PR #386, opened 2023-10-16
use debugPrintf instead of Serial.print* and fix format string

mar0ni:
Not sure why LittleFS usage reporting did not use the existing debugPrintf function. Also, the format string for usage was not correct and not value was logged. 

So, this fix uses debugPrint, fixes the format string, calculates the percentage for the usage and prints one line.


kjyv:
I think the reason was that at this point, there is no WiFi connection yet, so the logging over network won't work anyway.

mar0ni:
ok, got it. but in that case, doesn't make it even more sense to call debugPrintf and have the decision about to what channel (serial or wifi) to log in one central place rather than scattered across the code base?

kjyv:
Sure, in general it's not a bad idea to always call a single logging method. It was just to give an answer to your question as to why it was not used so far.

kjyv (approved):

PR #396, opened 2023-11-28
added ABP2 pressure sensor support

stefan23:
need to add to userConfig.h:
#define PRESSURE_VERBOSE 1 // if you want extra debug output
#define PRESSURESENSOR 2           // 0 = no pressure sensor connected, 1 = pressure sensor connected, 2 = ABP2 pressure sensor connected


LoQue90:
could you add the changes for the userConfig_sample.h in this PR, so the build actions won´t fail? :)
i tryed this yesterday, was working good (onyl checked mqtt output and serial)!

stefan23:
> could you add the changes for the userConfig_sample.h in this PR, so the build actions won´t fail? :) i tryed this yesterday, was working good (onyl checked mqtt output and serial)!

Sure. Sorry for the delay, I was more or less offline during Christmas break. Will the build actions understand it and try to build with `#define PRESSURESENSOR 2`?
And furthermore, I am still not fully familiar with GitHub and it seems that my subsequent commits after I filed the PR have been included into it. They are not fully completed yet and some code changes need adding the check for PRESSURESENSOR.

LoQue90:
No they will build with the plain userconfig_sample.
But right now the build actions can't test anything because with those missing pressure macros it can't build.
We should add more example config that will be build and checked.

fiendie (changes requested):

LoQue90 (approved):

PR #397, opened 2023-11-28
Water Sensor shown on display with icons

genkigenki:
Water Sensor, "water empty" icon shown in top right, on each display mode
In case the water sensor is not configured, water_full=true always, so no icon will ever be shown.


src/main.cpp
```diff
@@ -249,6 +250,7 @@ boolean setupDone = false;
 int backflushON = 0;             // 1 = backflush mode active
 int flushCycles = 0;             // number of active flush cycles
 int backflushState = 10;         // counter for state machine
+boolean waterFull = true;               // current state of water reservoir, true=full, false=empty
```
fiendie, line 253: The comment is a bit redundant, I think it's pretty clear from context what this variable is for.

src/main.cpp
```diff
@@ -2383,6 +2393,27 @@ void loopLED() {
     }
 }
 
+void loopWATER() {
+        if (digitalRead(PIN_WATERSENSOR) == HIGH) {
```
christian-hellwig, line 2397: No bounce time defined

src/main.cpp
```diff
@@ -2383,6 +2393,27 @@ void loopLED() {
...
+            //empty
+            if (waterFull == true)
+            {
+                debugPrintf("Water empty\n");
```
christian-hellwig, line 2401: logic error waterFull puts out water empty

src/main.cpp
```diff
@@ -2145,6 +2151,10 @@ void loop() {
         loopLED();
     }
 
+    if (WATER_SENS) {
+        loopWATER();
```
fiendie, line 2155: This should definitely not be called on every loop. This is just a waste of resources.

src/userConfig_sample.h
```diff
@@ -63,6 +63,7 @@ enum MACHINE {
 #define PINMODEVOLTAGESENSOR INPUT // Mode INPUT_PULLUP, INPUT or INPUT_PULLDOWN_16 (Only Pin 16)
 #define PRESSURESENSOR 0           // 0 = no pressure sensor connected, 1 = pressure sensor connected
 #define TEMP_LED 1                 // Blink status LED when temp is in range
+#define WATER_SENS 0               // 0 = no water sensor, 1 = water sensor XKC-Y25-NPN connected
```
christian-hellwig, line 66: see comment on digital read, config for npn or pnp missing

LoQue90: Could be defined like the switches 0 =no sensor 1= npn 2 = pnp

src/main.cpp
genkigenki: No its only printing if the "old state" is full, and the "new state" is empty. So the if statement ensures that only state changes trigger a print

src/main.cpp
genkigenki: debounced - in principle correct, this is the first version just for showing the icon, so i thought debouncing is not needed. But when testing with the serial output debouncing is however really not needed, water level not really changes, its very stable and may change when you push the water canister away from the sensor, but comes back once the water canister is back in place.

src/main.cpp
```diff
@@ -2383,6 +2393,27 @@ void loopLED() {
     }
 }
 
+void loopWATER() {
```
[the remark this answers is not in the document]
fiendie: 1. Then I fear you don't understand how a state machine works. Of course there would have to be state transitions where appropriate.
2. The water level is not the issue here. There can always be noise at the GPIO bin. You want the signal to be stable for a period before you do anything. I don't care if it works on your bench.

src/main.cpp
```diff
@@ -154,6 +154,7 @@ void setBDPIDTunings();
 void loopcalibrate();
 void looppid();
 void loopLED();
+void loopWATER();
```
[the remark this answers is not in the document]
genkigenki: fixed

src/main.cpp
genkigenki: fixed and added that without water sensor this will stay on "true"

src/userConfig_sample.h
genkigenki: Do you have a PNP to test? I have only NPN...

src/main.cpp
genkigenki: Checked again - no debounce needed at all

src/main.cpp
genkigenki: Called now every 100ms

src/main.cpp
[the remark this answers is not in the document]
genkigenki: 1. Hope I do. I think we need both: I would like to indicate "water empty" on the display e.g. also during a brew (or during other things like drawing water, steaming), not after the brew. That would need two MachineState, brew-water-full and brew-water-empty with a transition. To avoid this duplication, a separate Water-State is needed in my eyes.
But whent he machine is "locked" due to missing water, we need a MachineState "kWaterEmpty", with a clear indication that brewing is not possible at the moment. I simply did not implement this yet, I will do that once I used it for a while.
2. Currently, nothing is done except for the icon in the display. Will implement the debounce once the state chanbe is implemented.

src/main.cpp
genkigenki: Will debounce when implementing the state change (see comment below)

src/userConfig_sample.h
genkigenki: OK implemented

src/main.cpp
genkigenki: Debounce implemented, PNP  implemented

src/main.cpp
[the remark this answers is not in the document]
genkigenki: 2. Implemented.
For 1 - need to draw the state diagram to make sure it works. Will come later.

PR #402, opened 2023-12-03
RGB LEDs as state indicators with animations, cup illumination, etc

genkigenki:
LoopLED will be called every 20ms for smooth LED animations.
Number of LEDs and chip type is freely configurable.

    Heating to setpoint or Standby -> Default: switch all LEDs off
    Perfect temp -> Green, one LED in back of machine
    Brew/Steam -> Blue (only if configured)
    After coffee is brewed -> White light to show ready coffee for 15s
    Initialize, cold start Heat up --> Rainbow (only if configured)
    Error message: Red heartbeat (Water empty, sensor error, etc)


LoQue90:
This also adds all the stuff from your water icon pr.
This PR should only implement your led implementation.

LoQue90 (commented):
I would suggest STATUS_LED

PR #404, opened 2023-12-03
Web page proposal for Version 4.0.0 ESP32

Clemens-Toegel:
- Change API path from AsynWebServer
- Include new web page

As promised here is my redesign of the web page on the microcontroller.


kjyv:
This deletes all of the existing webinterface in favor of one minified js file - how is this supposed to be maintained? (or even reviewed)

Clemens-Toegel:
The `.editorconfig` is already in the repo, so it can be edited to your configuration, so others use the style you want I guess...

Okay, nice. Thanks for the input. That would be nice if it can be build automatically. I'm not very familiar with platform io, so I would definitely need your help there.
In the build everything is included, so the ESP does not need the internet for the webpage.

fiendie:
We added a CONTRIBUTING file as a reference for the code style. I think this is more flexible and allows for exceptions when they make sense.

fiendie:
@Clemens-Toegel: Are you still planning to publish the code? Otherwise I would close this PR.

fiendie:
After discussing this further, this is currently not a priority for the project. Feel free to open a new PR once the code is published and you have a detailed workflow for how to build and maintain the code.

Issue #416, opened 2023-12-09
Erratic behaviour of "Steam" switch when configured as push button


genkigenki:
Steam switch (in push button configuration) sometimes switches off by itself.

In main wird jeden looppid() aufgerufen (ohne debounce!):

void checkSteamON() {
[...] 
    // check digital GIPO
    if (digitalRead(PIN_STEAMSWITCH) == HIGH) {
        steamON = 1;
    }

und zusätzlich später jeden LOOP aus steamswitchvoid.h (mit debounce):

void checkSteamSwitch() {
[...]
        int reading = digitalRead(PIN_STEAMSWITCH);
[...]                        steamON = 1;

LoQue90:
https://github.com/rancilio-pid/clevercoffee/pull/430

Issue #413, opened 2023-12-09
Inonsistent Brew switch behaviour - likely debounce missing


genkigenki:
This code block has no function:

            if ((millis() - lastDebounceTimeBrewTrigger) > debounceDelayBrewTrigger) {
                // whatever the reading is at, it's been there for longer than the debounce
                // delay, so take it as the actual current state:

                // if the button state has changed:
                if (reading != buttonStateBrewTrigger) {
                    buttonStateBrewTrigger = reading;
                }
            }

as buttonStateBrewTrigger is not used anywhere else.
--> Brewswitchtrigger is not debounced.

This leads to issues in my machine.
Also the 5s delay was likely is introduced due to these issues ("Prellen")
Remove that as well

https://github.com/rancilio-pid/clevercoffee/assets/3080081/d1834d7d-1d32-4769-bb84-85508d9e67f9

LoQue90:
https://github.com/rancilio-pid/clevercoffee/pull/430

PR #417, opened 2023-12-10
 Bugfix: Inonsistent Brew & Steam switch behaviours #413 #414

genkigenki:
Brew button was not debounced, leading to erratic behaviour if your switches are not perfect.
5s delay removed, as its not needed (was likely introduced due to erratic behaviour due to bouncing switches)
Fixes https://github.com/rancilio-pid/clevercoffee/issues/413

Steam button configured as push button was twice checked and once not debounced. Removed check in main.cpp. Also migrated missing code to switchvoid file from main to ensure behaviour for switch users stays same.
Fixes https://github.com/rancilio-pid/clevercoffee/issues/416

### DIFF
--- a/src/brewvoid.h
+++ b/src/brewvoid.h
@@ -25,13 +25,9 @@ void checkbrewswitch() {
 
             if (reading != lastButtonStateBrew) {
                 // restart the debouncing timer
-                #if VERBOSE
-                    if(millis() - lastDebounceTimeBrewTrigger <= debounceDelayBrewTrigger) {
-                        debugPrintf("- No push detected, below debounce: %lu ms <= %lu ms\n", millis() - lastDebounceTimeBrewTrigger, debounceDelayBrewTrigger);
-                    }
-                #endif
                 lastDebounceTimeBrewTrigger = millis();
                 // set new button state 
+                lastButtonStateBrew = reading;
             }
             else if ((millis() - lastDebounceTimeBrewTrigger) > debounceDelayBrewTrigger) {
                 // whatever the reading is at, it's been there for longer than the debounce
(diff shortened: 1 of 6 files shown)


src/brewvoid.h
```diff
@@ -23,22 +23,21 @@ void checkbrewswitch() {
...
-            if ((millis() - lastDebounceTimeBrewTrigger) > debounceDelayBrewTrigger) {
+            else if ((millis() - lastDebounceTimeBrewTrigger) > debounceDelayBrewTrigger) {
                 // whatever the reading is at, it's been there for longer than the debounce
                 // delay, so take it as the actual current state:
-
-                // if the button state has changed:
-                if (reading != buttonStateBrewTrigger) {
-                    buttonStateBrewTrigger = reading;
+                
+                if (brewswitchTrigger != reading)
```
kjyv, line 36: Could you put the closing bracket inline with the if statement? It's the code style we generally use

kjyv (commented):
lgtm, although not sure about the 5 second delay. Could this be there for making the (software) brew detection work? The comment mentions that. So please test if it is really unnecessary

src/brewvoid.h
genkigenki: Can't be. Its the mechanical switch, converted to a push button:

Code is executed if:
#if BREWSWITCHTYPE == 2
        #if (PIN_BREWSWITCH > 0)

Config file:
#define BREWSWITCHTYPE 2           // 0 = no switch connected, 1 = normal switch, 2 = trigger switch

I tested with the brew switch, no issues at all.

kjyv: Ah yeah, the code path wouldn't make sense with software brew detection - the trigger switch is detected already. Then it should be fine

kjyv (approved):

LoQue90 (changes requested):
works worse than befor. dosen´t react do all pushes, brew start/stops not reliable
debounceDelayBrewTrigger set to 80 for test, did not help

kjyv:
Thanks for testing again. My review was not based on actually testing this as I don't have a push button installed in my machine, guess I should have added that caveat

LoQue90:
> Thanks for testing again. My review was not based on actually testing this as I don't have a push button installed in my machine, guess I should have added that caveat

don´t worry, the code is fine and works.
Problem is the debounce timing. Look like my stil of pushing that button is more a snap, which needs a faster debounce of 20ms.
Just wanted to check that :)

PR #436, opened 2024-01-06
Update shot weight after brew stopped to consider trickling after targetWeight - scaleDelayValue has been reached

FabianSperrle:
Currently, the shot stops when targetWeight - scaleDelayValue (default value: 2.5 grams) is reached. The current weight is stored as brewWeight. 
Then, the afterShot-Shottimer is displayed with the brewWeight.

At the same time, some coffee continues to trickle into the cup -- the whole point of the scaleDelayValue-logic. This is not reflected in the display.

This PR adds a new brewState to wait for two seconds of trickling, before officially stopping the shot.

### DIFF
--- a/src/brewvoid.h
+++ b/src/brewvoid.h
@@ -18,7 +18,8 @@ enum BrewState {
     kBrewRunning = 40,
     kWaitBrew = 41,
     kBrewFinished = 42,
-    kWaitBrewOff = 43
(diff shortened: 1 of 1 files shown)


LoQue90 (approved):

FabianSperrle:
Superseded by #450

PR #437, opened 2024-01-06
Add "flow rate" calculation by measuring what arrives in the cup

FabianSperrle:
- determine weight in the cup whenever we get a new weight measurement, divide by time since last measurement to get flow rate into the cup
- added a sliding window average over the last 10 measurements to get smoother output
- adjusted display template to show flow rate

### DIFF
--- a/src/brewvoid.h
+++ b/src/brewvoid.h
@@ -53,18 +53,23 @@ boolean brewPIDDisabled = false;                    // is PID disabled for delay
     boolean scaleCalibrationOn = 0;
     boolean scaleTareOn = 0;
     int shottimerCounter = 10 ;
-    float calibrationValue = SCALE_CALIBRATION_FACTOR;  // use calibration example to get value
     float weight = 0;                                   // value from HX711
     float weightPreBrew = 0;                            // value of scale before wrew started
     float weightBrew = 0;                               // weight value of brew
     float scaleDelayValue = 2.5;                        // value in gramm that takes still flows onto the scale after brew is stopped
     bool scaleFailure = false;
-    const unsigned long intervalWeight = 200;           // weight scale
+    const unsigned long intervalWeight = 50;           // weight scale
     unsigned long previousMillisScale;                  // initialisation at the end of init()
     HX711_ADC LoadCell(PIN_HXDAT, PIN_HXCLK);
     #if SCALE_TYPE == 0 
         HX711_ADC LoadCell2(PIN_HXDAT2, PIN_HXCLK);
     #endif 
+
+    // flow rate calculation
+    unsigned long prevFlowRateTime = 0;
+    float prevFlowRateWeight = 0.0;
+    float flowRate = 0.0;
+    float flowRateEmaAlpha = 0.05;
 #endif
 
 /**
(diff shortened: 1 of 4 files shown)


FabianSperrle:
In case this is well-received: would you also want flowrate (and weight) published via MQTT?

LoQue90:
I'm not into the scale, need the space to the portafilter for the cup, so I can't judge any scale related features.
But until now, anything that we measure or adjust is published to MQTT and is modified if possible through MQTT.
So I would say, go for. :)

FabianSperrle:
> I'm not into the scale, need the space to the portafilter for the cup, so I can't judge any scale related features. But until now, anything that we measure or adjust is published to MQTT and is modified if possible through MQTT. So I would say, go for. :)

Until I buy a bottomless portafilter I just screwed off the spout to gain some space, otherwise it would be painful, yes...

src/brewvoid.h
```diff
@@ -426,6 +435,44 @@ void brew() {
...
+                if (flowRateMeasurements < flowRateBufferSize) {
+                    flowRateMeasurements++;
+                }
+
+                // Calculate the average flow rate
+                float averageFlowRate = 0.0;
+                for (int i = 0; i < flowRateMeasurements; i++) {
+                    averageFlowRate += flowRateBuffer[i];
+                }
+                flowRate = averageFlowRate / flowRateBufferSize;
```
kjyv, line 465: This is basically a box car filter which is not a very good low pass filter. Maybe consider using an exponential moving average here as it has a much simpler implementation and faster response to recent changes. We already have an implementation in filterPressureValue() albeit with fixed parameters. You could make that method generic and use it here (e.g. with 0.01 and 0.99 as params).

FabianSperrle: I changed it to use EMA -- will determine potentially more fitting params with the next shots tomorrow ;) The calculation is so simple that I found it easier to do it in place rather than making the existing method generic...

src/brewscaleini.h
```diff
@@ -49,7 +49,13 @@ unsigned long previousMillistempanalogreading;      // ms for analogreading
     float weightBrew = 0;                               // weight value of brew
     float scaleDelayValue = 2.5;                        // value in gramm that takes still flows onto the scale after brew is stopped
     bool scaleFailure = false;
-    const unsigned long intervalWeight = 200;           // weight scale
+    const unsigned long intervalWeight = 50;           // weight scale
```
FabianSperrle, line 52: Increased the scale frequency -- why was it so low before? 
Potentially there could be a further change to only read the scale during shots to save some CPU cycles when idling?

src/scalevoid.h
```diff
@@ -80,8 +80,6 @@ void shottimerscale() {
             break;
 
         case 20:
-            weightBrew = weight - weightPreBrew;
-
```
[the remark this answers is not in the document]
LoQue90: Half done refactoring and moving stuff out of main?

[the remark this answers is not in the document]
FabianSperrle: Unless somebody remebers/knows why it should be here I think removing it here is good. Now that I looked at it again I would guess that the entire `shottimerscale()` could be removed.

FabianSperrle:
Superseded by #450

PR #445, opened 2024-01-22
Feature tests

fiendie:
Enable features in userConfig for automatic builds in GitHub Actions.


src/userConfig_sample.h
```diff
@@ -28,16 +28,27 @@ enum MACHINE {
 // Machine
 #define MACHINEID RancilioSilvia   // Machine type (see enum MACHINE)
 
+/* Features
+ *
+ * 0 = deactivated, 1 = activated
+ */
+#define FEATURE_HEATINGLOGO 1
```
kjyv, line 35: I like the rename but wouldn't it be nice to keep the on/off flag next to the options that relate to that feature?

fiendie: It seemed like a good idea to have the features all in one place but I'm not married to the idea :)

kjyv (approved):

PR #446, opened 2024-01-26
Go crazy with matrix builds

FabianSperrle:
maybe a bit much? :) but this is what I had in mind when @fiendie introduced the idea of testing more configurations. 

Obviously this will cost significant pipeline hours -- not sure how much of the quota is currently used in a typical month?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,20 +3,51 @@ name: Build
 on: [push, pull_request]
 
 jobs:
-    build:
-      runs-on: ubuntu-latest
-      steps:
-        - name: Checkout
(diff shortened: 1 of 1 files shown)


fiendie:
I actually don't have a clue how much of the quota we are actually using. The usage report says that we've used 0 of the 2000 free minutes this month. Which doesn't sound right but I highly doubt that we've come anywhere near those 2000 minutes.

kjyv (approved):
nice

fiendie:
> nice

It is nice, unfortunately it still doesn't run :D

fiendie:
This will be obsolete when #547 is merged

PR #450, opened 2024-01-27
Scale upgrades: Flow rate and trickle timer

FabianSperrle:
combine #436 and #437 into one PR.
Add fix for `SCALE2_CALIBRATION_FACTOR` initialization.

### DIFF
--- a/src/brewvoid.h
+++ b/src/brewvoid.h
@@ -18,7 +18,8 @@ enum BrewState {
     kBrewRunning = 40,
     kWaitBrew = 41,
     kBrewFinished = 42,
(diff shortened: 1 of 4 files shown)


FabianSperrle:
Even though it's not super necessary, I kinda miss the flow rate now that I updated to the latest master. 
Should I rebase this, maybe behind a feature flag, or just hack it in on my fork?

LoQue90:
I would say we could add this.
I'm only a bit concerned about the amount of stuff that will be shown on the display.

FabianSperrle:
Would need to be tested which features here are not yet present in master and what is still needed.

PR #453, opened 2024-02-06
Cleaned up project structure

fiendie:
* Created subfolders for hardware (sensors, switchs, etc.) and display (templates, bitmaps)
* Renamed some header files and classes in accordance with C/C++ naming conventions


src/PeriodicTrigger.h
```diff
@@ -11,8 +11,20 @@ class PeriodicTrigger {
  public:
     PeriodicTrigger(unsigned long millisec);
 
+    /**
+     * @brief Check if the interval period has elapsed
+     * @return Returns true untill the timer is reset
```
kjyv, line 16:
```suggestion
     * @return Returns true until the timer is reset
```

src/Storage.h
```diff
@@ -1,7 +1,7 @@
 /**
  * @file Storage.h
  *
- * @brief
+ * @brief Provides functions to manage data in a non-volatile memory
```
kjyv, line 4:
```suggestion
 * @brief Provides functions to manage data in non-volatile memory
```

src/display/displayTemplateTemponly.h
```diff
@@ -1,5 +1,5 @@
 /**
- * @file Displaytemplatetemponly.h
+ * @file displayTemplateTemponly.h
```
kjyv, line 2:
```suggestion
 * @file displayTemplateTempOnly.h
```

src/embeddedWebserver.h
```diff
@@ -161,15 +161,15 @@ void paramToJson(String name, editable_t &e, DynamicJsonDocument &doc) {
...
+// hash strings at compile time to use in switch statement
+// (from https://stackoverflow.com/questions/2111667/compile-time-string-hashing)
 constexpr unsigned int str2int(const char* str, int h = 0) {
     return !str[h] ? 5381 : (str2int(str, h+1) * 33) ^ str[h];
 }
 
 String getHeader(String varName) {
-    //TODO: actually put the references libs on local file system again (only when using ESP32 which has more flash mem,
-    //but also make sure web server can handle this many concurrent requests (might crash)
+    // TODO: actually put the references libs on local file system again (only when using ESP32 which has more flash mem,
```
kjyv, line 171: I think this comment can go, all files are in the filesystem.
Btw, since when is NOINTERNET 1 by default?

src/DebugSerial.cpp
```diff
@@ -1,11 +1,11 @@
 /**
- * @file debugSerial.cpp
+ * @file DebugSerial.cpp
```
kjyv, line 2: why are some file names capitalized and others aren't?

fiendie: C++ Classes and their header files are usually capitalised while C header files are not. 
The only grey area is SysPara.h because it doesn't have a separate implementation file.

src/embeddedWebserver.h
fiendie: Wasn't me! ;D

src/Storage.h
fiendie: I just copy-pasted this from the implementation file but I'll change it in both files.

src/DebugSerial.cpp
kjyv: I was thinking about that but e.g. in this file, no classes are defined

src/embeddedWebserver.h
kjyv: I guess it's not a bad default for consistency, just slower

src/DebugSerial.cpp
fiendie: Fair enough! There shouldn't be a separate implementation file in that case, however.

kjyv (approved):

src/main.cpp
```diff
@@ -317,7 +317,7 @@ int steamFirstON = 0;
 double aggKd = aggTv * aggKp;
 
 // Timer - ISR for PID calculation and heat relay output
-#include "ISR.h"
+#include "isr.h"
```
fiendie, line 320: Meh, isr.h is included twice.

LoQue90 (approved):

PR #458, opened 2024-02-12
Add classes for handling GPIO Pins, Switches, Relays, Temp Sensors and LEDs

fiendie:
Still a work in progress but this is the first draft of a hardware abstraction layer. Thoughts, comments, criticism... :)


src/hardware/GPIOPin.h
```diff
@@ -0,0 +1,33 @@
...
+    GPIO_INPUT_ANALOG
+};
+
+class GPIOPin {
+    public:
+        GPIOPin(int pinNumber, PinType pinType);
+
+        void write(bool value);
+        void setPinType(PinType pinType);
+        int read();
```
simonspa, line 26: You could mark methods not changing the object as `const`:

```suggestion
        int read() const;
```

(same for many others) This makes it a bit more safe in cases where you *do not want to* alter anything. Then you can pass `const GPIOPin` and would get a compiler error if the code would try to ca.. e.g. `setPinType()`

src/hardware/IOSwitch.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+    if (reading != lastState) {
+        lastDebounceTime = currentMillis;
+    }
+
+    if ((currentMillis - lastDebounceTime) > debounceDelay) {
+        if (reading != currentState) {
+            currentState = reading;
+
+            if (currentState == LOW) {
```
simonspa, line 35: This logic seems flawed to me, the condition will never occur since you *just* set `currentState = reading`.

src/hardware/Relay.h
```diff
@@ -0,0 +1,26 @@
+/**
+ * @file Relay.h
+ * 
+ * @brief A relay connected to a GPIO pin
+ */
+#pragma once
+
+class GPIOPin;
+
+enum RelayTriggerType {
```
simonspa, line 10: Really just a matter of taste but I tend to define these inside the class:

```cpp
class Relay {
public:
  enum TriggerType {
    LOW_TRIGGER,
    HIGH_TRIGGER,
  };

  Relay(/*...*/);
  /* ... */
private:
  /* ... */
}
```

That way you don't have the doubling of `Relay` in the name. Inside the class you refer to it as `TriggerType` and if you need it outside, it's `Relay::TriggerType`.

src/hardware/TempSensorDallas.cpp
```diff
@@ -0,0 +1,27 @@
...
+ * 
+ * @brief Handler for Dallas DS18B20 temperature sensor
+ */
+
+#include "TempSensorDallas.h"
+#include <DallasTemperature.h> 
+
+OneWire* oneWire;
+DallasTemperature* dallasSensor;
+DeviceAddress sensorDeviceAddress;
```
simonspa, line 12: If they are only used by this class I would move them into the class as private members. In case this needs to be shared between multiple `TempSensorDallas` instances (which is not the case here) you could then mark them as `static`.

src/hardware/TempSensorTSIC.cpp
```diff
@@ -0,0 +1,22 @@
...
+ * 
+ * @brief Handler for TSIC 306 temperature sensor
+ */
+
+#include "TempSensorTSIC.h"
+#include <ZACwire.h>
+
+#define MAX_CHANGERATE 15
+
+ZACwire* tsicSensor;
```
simonspa, line 12: Also here, better as class member

src/main.cpp
```diff
@@ -2223,19 +2202,14 @@ void setup() {
...
-        sensors.setResolution(sensorDeviceAddress, 10);
-        sensors.requestTemperatures();
-        temperature = sensors.getTempCByIndex(0);
-    #endif
+    if (TEMP_SENSOR == 1) {
+        tempSensor = new TempSensorDallas(PIN_TEMPSENSOR);
+    }
+    else if (TEMP_SENSOR == 2) {
+        tempSensor = new TempSensorTSIC(PIN_TEMPSENSOR);
+    }
```
simonspa, line 2210: We know this at compile-time. I would directly set this in the top of the file along with the rest of the hardware. That way you don;t run the risk that this code is somehow not reached and you are operating with a `nullptr` whenever calling `tempSensor->something()` and crash the controller.

simonspa (commented):
Huge step forward in readability and maintainability from my perspective. Cudos @fiendie !

src/main.cpp
simonspa: Optimally we would not even include the Dallas sensor lib in the build, so only required code ends up on the device.

src/hardware/IOSwitch.cpp
simonspa: I misread. This is fine.

src/userConfig_sample.h
```diff
@@ -93,5 +98,8 @@ enum MACHINE {
 // PID Parameters (not yet in Web interface)
 #define EMA_FACTOR 0.6             // Smoothing of input that is used for Tv (derivative component of PID). Smaller means less smoothing but also less delay, 0 means no filtering
 
-#define TEMPSENSOR 2               // Temp sensor type: 1 = DS18B20, 2 = TSIC306
+#define TEMP_SENSOR 2               // Temp sensor type: 1 = DS18B20, 2 = TSIC306
+
+// Log level for serial console
+#define LOGLEVEL Logger::Level::INFO
```
kjyv, line 104: Should still include a comment for other options here (this file still kind of an end user config file, so most are likely not aware of log levels, also not which are relevant here)

lib/Logger/Logger.cpp
```diff
@@ -0,0 +1,134 @@
...
+    switch(lvl) {
+    case Level::TRACE:   return "  TRACE";
+    case Level::DEBUG:   return "  DEBUG";
+    case Level::INFO:    return "   INFO";
+    case Level::WARNING: return "WARNING";
+    case Level::ERROR:   return "  ERROR";
+    case Level::FATAL:   return "  FATAL";
+    default:             return " SILENT";
+    }
+}
```
kjyv, line 134:
```suggestion
String Logger::get_level_identifier(Logger::Level lvl) {
    switch(lvl) {
        case Level::TRACE:   return "  TRACE";
        case Level::DEBUG:   return "  DEBUG";
        case Level::INFO:    return "   INFO";
        case Level::WARNING: return "WARNING";
        case Level::ERROR:   return "  ERROR";
        case Level::FATAL:   return "  FATAL";
        default:             return " SILENT";
    }
}
```

src/brewHandler.h
```diff
@@ -31,16 +31,13 @@ enum BackflushState {
...
-int lastStateBrewTrigger;                        // the last valid reading from the input pin (debounced)
-unsigned long lastDebounceTimeBrewTrigger = 0;  // the last time the output pin was toggled
-unsigned long debounceDelayBrewTrigger = 20;    // >20ms when switches are "flicked"
-unsigned long brewSwitchTriggerMillis = 0;
-const unsigned long brewTriggerLongPress = 500;     // time in ms until brew trigger will be interpreted as manual brewing
-int brewSwitchTriggerCase = 10;
+BrewState currBrewState = kBrewIdle;
+
+uint8_t currStateBrewSwitch = LOW;
+uint8_t currBrewSwitchStateMomentary = LOW;
```
kjyv, line 37: These variables sound very similar, but it's not clear from their names (especially the second one) what they're for

kjyv: Why do we need two variables anyway though, the switch can always only have one type I guess

src/brewHandler.h
```diff
@@ -72,89 +69,75 @@ boolean brewPIDDisabled = false;                    // is PID disabled for delay
...
+        currStateBrewSwitch = brewSwitchReading;
+    }
+    else if (BREWSWITCH_TYPE == Switch::MOMENTARY) {
+        if (currBrewSwitchStateMomentary != brewSwitchReading) {
+            currBrewSwitchStateMomentary = brewSwitchReading;
+        }
+        
         // Convert trigger signal to brew switch state
-        switch (brewSwitchTriggerCase) {
+        switch (brewSwitchCase) {
```
kjyv, line 83: This variable might be better called something like `brewSwitchState` as this is a state machine and the variable holds the state. This might get confusing with the switch state variables however, so maybe these should not include state but rather value?

src/display/displayCommon.h
```diff
@@ -238,10 +238,10 @@ void displayLogo(String displaymessagetext, String displaymessagetext2) {
  * @brief display shot timer
  */
 void displayShottimer(void) {
-    if ((machineState == kBrew || brewSwitchTriggerCase == 31) && FEATURE_SHOTTIMER == 1 && SHOTTIMER_TYPE == 1) {
+    if ((machineState == kBrew || brewSwitchCase == 31) && FEATURE_SHOTTIMER == 1 && SHOTTIMER_TYPE == 1) {
```
kjyv, line 241: Not really the focus of this PR but can we also replace these magic numbers with a proper enum for the states like for the machineState?

src/display/displayTemplateMinimal.h
```diff
@@ -35,7 +35,7 @@ void printScreen() {
             }
 
             // Draw temp, blink if TEMP_LED is not enabled
```
kjyv, line 37:
```suggestion
            // Draw temp, blink if STATUS_LED is not enabled
```

src/hardware/Relay.h
```diff
@@ -0,0 +1,55 @@
...
+// Forward declaration of GPIOPin class
+class GPIOPin;
+
+/**
+ * @class Relay control class
+ * @brief This class provides control for relay switches
+ */
+class Relay {
+    public:
+
```
kjyv, line 17:
```suggestion
```

src/hardware/StandardLED.cpp
```diff
@@ -0,0 +1,28 @@
...
+ * @file StandardLED.cpp
+ * 
+ * @brief An LED connected to a GPIO pin
+ */
+
+#include "GPIOPin.h"
+#include "StandardLED.h"
+
+StandardLED::StandardLED(GPIOPin& gpioInstance) : gpio(gpioInstance) {
+
```
kjyv, line 11:
```suggestion
```

src/hardware/TempSensor.h
```diff
@@ -0,0 +1,32 @@
...
+
+class TempSensor {
+    public:
+        /**
+         * @brief Returns the current temperature
+         * @details Requests sampling from attached sensor and returns reading. This method is purely virtual and must be
+         *          implemented by every child class.
+         * @return Temperature in degrees Celsius
+         */
+        virtual float getTempinCelsius() const = 0;
```
kjyv, line 17:
```suggestion
        virtual float getTempInCelsius() const = 0;
```

src/main.cpp
```diff
@@ -443,6 +461,10 @@ const unsigned long intervalDisplay = 500;
     #endif
 #endif
 
+#include "powerHandler.h"
+#include "steamHandler.h"
+#include "hardware/brewScale.h"
```
kjyv, line 466: We could consider renaming brewScale to scaleHandler

kjyv (commented):
Nice work, just  a few comments, otherwise approved

src/main.cpp
LoQue90: therer was a short idea of adding a acaia bluetooth scale, but way in the future :)
better naming schema would be nice

src/brewHandler.h
```diff
@@ -72,89 +69,75 @@ boolean brewPIDDisabled = false;                    // is PID disabled for delay
...
-                    brewSwitchTriggerCase = 31;
-                    digitalWrite(PIN_VALVE, relayOn);
-                    digitalWrite(PIN_PUMP, relayOn);
+                // Button more than brewSwitchMomentaryLongPress pushed
+                if (currBrewSwitchStateMomentary == HIGH && brewSwitch->longPressDetected() && machineState != kWaterEmpty) {
+                    brewSwitchCase = 31;
+                    valveRelay.on();
+                    pumpRelay.on();
                     startingTime = millis();
+                    LOG(DEBUG, "brewSwitchCase 20: Manual Trigger - flushing");
```
LoQue90, line 106: this should be log level INFO, like all brewing states ( Prefinfusion, pre pause, brew
maybe we should move the action (pumpe and vavle) into the brewstatemashine and not trigger this out of the brewswitch state mashine?

src/embeddedWebserver.h
```diff
@@ -494,7 +494,7 @@ void serverSetup() {
 
     server.begin();
 
-    debugPrintln(("Server started at " + WiFi.localIP().toString()).c_str());
+    LOG(DEBUG, ("Server started at " + WiFi.localIP().toString()).c_str());
```
LoQue90, line 497: LOG(INFO,

src/hardware/GPIOPin.cpp
```diff
@@ -0,0 +1,57 @@
...
+            break;
+        case IN_PULLUP:
+            pinMode(pin, INPUT_PULLUP);
+            digitalWrite(pin, HIGH);
+            break;
+        case IN_PULLDOWN:
+            pinMode(pin, INPUT_PULLDOWN);
+            digitalWrite(pin, LOW);
+            break;
+        case IN_HARDWARE:
```
LoQue90, line 50: maybe a comment: hardware= externall pull resistor?

LoQue90: nvm, enum has the comment!

src/main.cpp
```diff
@@ -622,15 +608,15 @@ void initOfflineMode() {
         displayMessage("", "", "", "", "Begin Fallback,", "No Wifi");
     #endif
 
-    debugPrintln("Start offline mode with eeprom values, no wifi :(");
+    LOG(DEBUG, "Start offline mode with eeprom values, no wifi :(");
```
LoQue90, line 611: log level INFO?

src/main.cpp
```diff
@@ -622,15 +608,15 @@ void initOfflineMode() {
...
     offlineMode = 1;
 
     if (readSysParamsFromStorage() != 0) {
         #if OLED_DISPLAY != 0
             displayMessage("", "", "", "", "No eeprom,", "Values");
         #endif
 
-        debugPrintln(
+        LOG(DEBUG,
```
LoQue90, line 619: log level INFO?

src/main.cpp
```diff
@@ -654,7 +640,7 @@ void checkWifi() {
             if (statusTemp != WL_CONNECTED) {  // check WiFi connection status
                 lastWifiConnectionAttempt = millis();
                 wifiReconnects++;
-                debugPrintf("Attempting WIFI (re-)connection: %i\n", wifiReconnects);
+                LOGF(DEBUG, "Attempting WIFI (re-)connection: %i", wifiReconnects);
```
LoQue90, line 643: log level INFO?

src/main.cpp
```diff
@@ -1155,7 +1141,7 @@ void handleMachineState() {
             brewDetection();
 
             if (millis() - lastBrewTimeMillis > SHOTTIMERDISPLAYDELAY) {
-                debugPrintf("Shot time: %4.1f s\n", lastBrewTime / 1000);
+                LOGF(DEBUG, "Shot time: %4.1f s", lastBrewTime / 1000);
```
LoQue90, line 1144: log level INFO?

src/brewHandler.h
simonspa: I had them at `DEBUG` because printing out enum values of a switch case felt rather technical. Can we leave this for post-merging to decide or improve?

lib/Logger/Logger.cpp
simonspa: Fixed this manually. I am preparing a follow-up PR after this one is merged with `pre-commit` config and `clang-format` such that we always get consistent code formatting and don't have to look for this in reviews.

src/brewHandler.h
LoQue90: I gonna put that on the roadmap, but I would like to get this output back as soon as possible. Without there is no log feedback for the long press brewswitch and flushing

src/hardware/StandardLED.cpp
```diff
@@ -7,9 +7,7 @@
 #include "GPIOPin.h"
 #include "StandardLED.h"
 
-StandardLED::StandardLED(GPIOPin& gpioInstance) : gpio(gpioInstance) {
-
-}
+StandardLED::StandardLED(GPIOPin& gpioInstance) : gpio(gpioInstance) {}
```
LoQue90, line 10: The curly went one line to far up. 😆

simonspa: You sure? :D

I'd suggest let's leave the individual styling comments for after this MR, see my comment above: I am suggesting a proper `clang-format` setup with `pre-commit` and GH actions - then we don't have to bother with nuisances like bracket positions anymore

src/main.cpp
fiendie: The "brew" prefix was supposed to differentiate it from stuff related to scaling as opposed to weighing things. But maybe it's clear from context :)

src/hardware/StandardLED.cpp
LoQue90: As long as we don't want to chance our code style I'm sure :p

src/brewHandler.h
fiendie: Renamed to brewSwitchState

src/brewHandler.h
fiendie: To be perfectly honest, I didn't change much about the logic of how the momentary switch is handled. My main goal was to handle the debouncing in the Switch class and make the variable names a bit more consistent/less confusing.

We should refactor that in a followup PR because it's working at the moment and I don't want to get he HAL into master as soon as possible.

src/brewHandler.h
fiendie: See above

src/display/displayCommon.h
fiendie: I thought about it for a few minutes and couldn't come up with good names. Suggestions welcome :D

src/brewHandler.h
LoQue90: Brewswitch state mashine and brew state mashine should get a refactore, yes
But as you said, in another PR. I did put it on the roadmap.

src/userConfig_sample.h
```diff
@@ -52,20 +51,26 @@ enum MACHINE {
...
+#define ONLYPID 1                                // 0 = PID and preinfusion, 1 = Only PID
+#define ONLYPIDSCALE 0                           // 0 = off , 1 = OnlyPID with Scale
+#define BREWMODE 1                               // 1 = Brew by time (with preinfusion); 2 = Brew by weight (from scale)
+#define FEATURE_BREWDETECTION 1                  // 0 = deactivated, 1 = activated
+#define BREWDETECTION_TYPE 1                     // 1 = Software (Onlypid 1), 2 = Hardware (Onlypid 0), 3 = optocoupler for Only PID
+#define FEATURE_POWERSWITCH 0                    // 0 = deactivated, 1 = activated
+#define POWERSWITCH_TYPE Switch::TOGGLE          // Switch::TOGGLE or Switch::MOMENTARY (trigger)
+#define FEATURE_BREWSWITCH 0                     // 0 = deactivated, 1 = activated
+#define BREWSWITCH_TYPE Switch::TOGGLE           // Switch::TOGGLE or Switch::MOMENTARY (trigger)
+#define FEATURE_STEAMSWITCH 1                    // 0 = deactivated, 1 = activated
```
LoQue90, line 63: #define FEATURE_STEAMSWITCH 0

src/brewHandler.h
kjyv: Sure, my second comment was a note on the side. But the variable names could be improved, they are part of the PR.

LoQue90:
@kjyv I added the enum for the brewswitch state mashine, if you are happy with that, I would merge the PR :)

src/brewHandler.h
```diff
@@ -69,92 +74,75 @@ boolean brewPIDDisabled = false;                    // is PID disabled for delay
...
 
-            case 31:
-                // Stop Manual brewing, button goes low:
-                if (brewSwitchTrigger == LOW && brewSwitch == LOW) {
-                    brewSwitchTriggerCase = 40;
-                    debugPrintln("brewswitchTriggerCase 31: Manual Trigger - brewing stop");
-                    digitalWrite(PIN_VALVE, relayOff);
-                    digitalWrite(PIN_PUMP, relayOff);
+            case kBrewSwitchFlushOff:
+                // Brew switch got releaased - stop flushing
```
kjyv, line 128:
```suggestion
                // Brew switch got released - stop flushing
```

src/display/displayTemplateMinimal.h
```diff
@@ -15,7 +15,7 @@ void printScreen() {
         ((machineState == kBrew || machineState == kShotTimerAfterBrew) && FEATURE_SHOTTIMER == 0) ||
         machineState == kCoolDown || ((machineState == kColdStart) && FEATURE_HEATINGLOGO == 0) || 
         ((machineState == kPidOffline) && FEATURE_OFFLINELOGO == 0))
-        && (brewSwitchTriggerCase != 31))
+        && (brewSwitchState != 31))
```
kjyv, line 18:
```suggestion
        && (brewSwitchState != kBrewSwitchFlushOff))
```

src/display/displayTemplateScale.h
```diff
@@ -15,7 +15,7 @@ void printScreen() {
         ((machineState == kBrew || machineState == kShotTimerAfterBrew) && FEATURE_SHOTTIMER == 0) ||  // shottimer == 0, auch Bezug anzeigen
         machineState == kCoolDown || ((machineState == kInit || machineState == kColdStart) && FEATURE_HEATINGLOGO == 0) ||
         ((machineState == kPidOffline) && FEATURE_OFFLINELOGO == 0))
-        && (brewSwitchTriggerCase != 31))
+        && (brewSwitchState != 31))
```
kjyv, line 18:
```suggestion
        && (brewSwitchState != kBrewSwitchFlushOff))
```

src/display/displayTemplateTempOnly.h
```diff
@@ -19,14 +19,14 @@ void printScreen() {
         ((machineState == kBrew || machineState == kShotTimerAfterBrew) && FEATURE_SHOTTIMER == 0) ||  // shottimer == 0, auch Bezug anzeigen
         machineState == kCoolDown || ((machineState == kInit || machineState == kColdStart) && FEATURE_HEATINGLOGO == 0) ||
         ((machineState == kPidOffline) && FEATURE_OFFLINELOGO == 0))
-        && (brewSwitchTriggerCase != 31))
+        && (brewSwitchState != 31))
```
kjyv, line 22:
```suggestion
        && (brewSwitchState != kBrewSwitchFlushOff))
```

src/display/displayTemplateUpright.h
```diff
@@ -15,7 +15,7 @@ void printScreen() {
         ((machineState == kBrew || machineState == kShotTimerAfterBrew) && FEATURE_SHOTTIMER == 0) ||  // shottimer == 0, auch Bezug anzeigen
         machineState == kCoolDown || ((machineState == kInit || machineState == kColdStart) && FEATURE_HEATINGLOGO == 0) ||
         ((machineState == kPidOffline) && FEATURE_OFFLINELOGO == 0))
-        && (brewSwitchTriggerCase != 31))
+        && (brewSwitchState != 31))
```
kjyv, line 18:
```suggestion
        && (brewSwitchState != kBrewSwitchFlushOff))
```

src/brewHandler.h
kjyv: I think this would be an easy change that could be done before the merge

src/display/displayTemplateMinimal.h
LoQue90: How did I missed them...

LoQue90 (approved):

simonspa:
🎉

PR #465, opened 2024-03-02
make sure the shot timer also shows when still in kAtSetpoint

kjyv:
The brew detection got removed recently from the kAtSetpoint state which is 1° below the setpoint but it seems it should also be run there. In this state, the heating logo is already disabled so the shot timer should also work as in kBrew. The display looks the same in both cases so this was a bit confusing.


LoQue90:
just wanted to do a rebase and merge, but removing FEATURE_SHOTTIMER = 1 will bring fullscreen shottimer also when FEATURE_SHOTTIMER = 0 is set. Or do i miss something?

LoQue90 (approved):

PR #467, opened 2024-03-03
Turn off display 10 minutes after entering standby mode

fiendie:


src/standby.h
```diff
@@ -30,12 +34,31 @@ void updateStandbyTimer(void) {
...
+        else {
+            standbyModeRemainingTimeDisplayOffMillis = 0;
+            LOG(INFO, "Turning off display...");
         }
     }
 }
 
 void resetStandbyTimer(void) {
-    standbyModeRemainingTimeMillis = standbyModeTime * 60 * 10000;
+    standbyModeRemainingTimeMillis = standbyModeTime * 60 * 1000;
```
kjyv, line 60: How does change existing settings, will everything be x10 faster?

kjyv (approved):

src/standby.h
fiendie: Can't even remember making that change 😄
Definitely doesn't make any sense...

kjyv: I guess no one noticed that the reset always made the next standby 10x as long ;)

PR #472, opened 2024-03-05
Adressable LEDs (like WS2811) in HAL

genkigenki:
Added adressable LEDs for the HAL.
Status / brew LEDs are automatically assigned first, remainder is assigned to context-LEDs.
All based on LED class: One RGBLEDstrip class provides access to WS2811/WS2812/...  Additional classes for single-LEDs and multiple-LEDs. Cleanup of LED code in main.cpp

Detail:
LoopLED will be called every 80ms for relatively smooth LED animations.
Number of LEDs and chip type are configurable.

PID-on (not standby) -> low brightness white
Perfect temp -> dark green
Brew/Steam/Backflush -> white
Initialize, cold start Heat up -> Colors
Error message -> Red heartbeat (Water empty, sensor error, etc)
Steam-heat-up -> dark yellow

### DIFF
--- a/src/hardware/RGBLEDstrip.cpp
+++ b/src/hardware/RGBLEDstrip.cpp
@@ -1,0 +1,105 @@
+/**
+ * @file RGBLEDstrip.cpp
+ * 
+ * @brief A strip with adressable LEDs connected to a GPIO pin
+ */
+
+#include "RGBLEDstrip.h"
+#include "chipsets.h"
+
+RGBLEDstrip::RGBLEDstrip(int numLEDs, Type type) 
+    : ledType(type),
+      numLEDs(numLEDs), 
+      numAssignedLEDs(0) {
+
+    leds = new CRGB[numLEDs];
+
+    switch(type) {
+        case LED::WS2811: 
+            FastLED.addLeds<::WS2811, PIN_STATUSLED>(leds, numLEDs);  
+            break;
+
+        case LED::WS2812: 
+            FastLED.addLeds<::WS2812, PIN_STATUSLED>(leds, numLEDs);             
+            break;
+
+        case LED::WS2812B: 
+            FastLED.addLeds<::WS2812B, PIN_STATUSLED>(leds, numLEDs);
+            break;
+    }
+
+    turnOff();
+}
+
+RGBLEDstrip::~RGBLEDstrip() {
+    delete[] leds;
+}
+
+void RGBLEDstrip::turnOn() {
+    setBrightness(255); 
+}
+
+void RGBLEDstrip::turnOff() {
+    setBrightness(0); 
+}
+
+void RGBLEDstrip::setColor(int hue, int saturation, int value) {
+    setLEDRangeColor(0, numLEDs-1, hue, saturation, value);
+}
+
+void RGBLEDstrip::setBrightness(int value) {
+    setColor(0, 0, value);
+}
+
+void RGBLEDstrip::setLEDindexColor(int index, int hue, int saturation, int value) {
+    setLEDRangeColor(index, index + 1, hue, saturation, value);
+}
+
+void RGBLEDstrip::prepareLEDRangeColor(int start, int endPlusOne, int hue, int saturation, int value) {
+    if(start < 0 || endPlusOne > numLEDs) 
+        return; 
+    
+    for(int i = start; i < endPlusOne; i++) { 
+        leds[i].setHSV(hue, saturation, value);
+    };    
+}
+
+void RGBLEDstrip::setLEDRangeColor(int start, int endPlusOne, int hue, int saturation, int value) {
+    prepareLEDRangeColor(start, endPlusOne, hue, saturation, value);
+    FastLED.show();
+}
+
+void RGBLEDstrip::setLEDRangeRainbow(int start, int endPlusOne, int initialHue, int deltaHue) {
+    if(start < 0 || endPlusOne > numLEDs) return; 
+    CHSV hsv;
+    hsv.hue = initialHue;
+    hsv.val = 255;
+    hsv.sat = 240;
+    for( int i = 0; i < endPlusOne; ++i) {
+        leds[i] = hsv;
+        hsv.hue += deltaHue;
+    }
+    FastLED.show();
+}
+
+void RGBLEDstrip::setLEDRangeColorTwoValues(int start, int endPlusOne, int hue, int saturation, int value1, int value2) {
+    if(start < 0 || endPlusOne > numLEDs) return; 
+    int numFirstHalf = ( endPlusOne - start ) / 2;
+    int idSecondHalf = start + numFirstHalf;
+    prepareLEDRangeColor(start, idSecondHalf, hue, saturation, value1);
+    prepareLEDRangeColor(idSecondHalf, endPlusOne, hue, saturation, value2);
+    FastLED.show();
+}
+
+int RGBLEDstrip::assignLED() {
+    numAssignedLEDs++;
+    return numAssignedLEDs-1;
+}
+
+int RGBLEDstrip::getNumAssignedLEDs() {
+    return numAssignedLEDs;
+}
+
+int RGBLEDstrip::getNumLEDs() {
+    return numLEDs;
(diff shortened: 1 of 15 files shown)


src/main.cpp
```diff
@@ -2501,21 +2530,90 @@ void looppid() {
...
+                brewLed->turnOn();
+            }
+            else {
+                brewLed->turnOff();
+            }
         }
-        else {
-            brewLed->turnOff();
+
+        if (FEATURE_CONTEXT_LED) {
```
fiendie, line 2553: I suspect you're trying to get around having to write if (FEATURE_CONTEXT_LED) in every case-statement but duplicating the whole state machine seems infinitely worse.

src/main.cpp
```diff
@@ -2501,21 +2530,90 @@ void looppid() {
...
+
+            // Color cycle variable
+            cycleLED++;
+            if (cycleLED >= 255) {
+                cycleLED = 0;
+            }
+
+            switch (machineState) {
+                case kPidNormal:
```
fiendie, line 2562: Besides, why even handle cases where nothing being done?

src/hardware/RgbMultipleLED.cpp
```diff
@@ -0,0 +1,47 @@
...
+void RgbMultipleLED::setColor(int hue, int saturation, int value) {
+    ledStrip.setLEDRangeColor(startLEDid, numLEDs, hue, saturation, value);
+}
+
+void RgbMultipleLED::setRainbow(int initialHue, int deltaHue) {
+    ledStrip.setLEDRangeRainbow(startLEDid, numLEDs, initialHue, deltaHue);
+}
+
+void RgbMultipleLED::setColorTwoValues(int hue, int saturation, int value1, int value2) {
+    //ledStrip.setColor(hue, 255, 255);
```
fiendie, line 45: Please remove commented out code

src/hardware/RgbMultipleLED.cpp
```diff
@@ -0,0 +1,47 @@
...
+    : ledStrip(ledStrip),
+      startLEDid (0),
+      numLEDs (0) {
+    // The constructor saves the reference to the RGBLED object.
+
+    startLEDid = ledStrip.getNumAssignedLEDs(); 
+    numLEDs = ledStrip.getNumLEDs(); 
+
+    //TODO
+    LOGF(WARNING,
```
fiendie, line 20: Remove line break

src/hardware/RgbMultipleLED.cpp
```diff
@@ -0,0 +1,47 @@
...
+RgbMultipleLED::RgbMultipleLED(RGBLEDstrip& ledStrip) 
+    : ledStrip(ledStrip),
+      startLEDid (0),
+      numLEDs (0) {
+    // The constructor saves the reference to the RGBLED object.
+
+    startLEDid = ledStrip.getNumAssignedLEDs(); 
+    numLEDs = ledStrip.getNumLEDs(); 
+
+    //TODO
```
fiendie, line 19: What's left to do?

src/hardware/RGBLEDstrip.cpp
```diff
@@ -0,0 +1,120 @@
...
+    return numAssignedLEDs;
+
+    //TODO  
+    LOGF(DEBUG, "RGBLEDstrip::getNumAssignedLEDs numAssignedLEDs = %i", numAssignedLEDs);    
+}
+
+int RGBLEDstrip::getNumLEDs() {
+    return numLEDs;
+    
+        LOGF(DEBUG, "RGBLEDstrip::getNumLEDs numLEDs = %i", numLEDs);
```
fiendie, line 119: This line is never called

src/hardware/RGBLEDstrip.cpp
```diff
@@ -0,0 +1,120 @@
...
+    prepareLEDRangeColor(start, idSecondHalf, hue, saturation, value1);
+    prepareLEDRangeColor(idSecondHalf, endPlusOne, hue, saturation, value2);
+    FastLED.show();
+}
+
+int RGBLEDstrip::assignLED() {
+    numAssignedLEDs++;
+    return numAssignedLEDs-1;
+
+        //TODO
```
fiendie, line 105: What's left to do?

src/hardware/RGBLEDstrip.cpp
```diff
@@ -0,0 +1,120 @@
...
+    return numAssignedLEDs-1;
+
+        //TODO  
+    LOGF(DEBUG, "RGBLEDstrip::assignLED numAssignedLEDs = %i", numAssignedLEDs);    
+}
+
+int RGBLEDstrip::getNumAssignedLEDs() {
+    return numAssignedLEDs;
+
+    //TODO
```
fiendie, line 112: This line is never called and it's still not clear what the TODO is.

src/hardware/RgbSingleLED.cpp
```diff
@@ -0,0 +1,35 @@
...
+ * 
+ * @brief Single adressable LED, like status or brew LED
+ */
+
+#include "RgbSingleLED.h"
+
+RgbSingleLED::RgbSingleLED(RGBLEDstrip& ledStrip) 
+    : ledStrip(ledStrip),
+      LEDid (0) {
+    // The constructor saves the reference to the RGBLED object.
```
fiendie, line 12: See above.

src/hardware/RgbSingleLED.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+void RgbSingleLED::turnOn() {
+    // Set the first LED to a specific color to indicate "on" status.
+    // For example, setting it to a bright white color using HSV.
+    setBrightness(255); // Set the first LED to 100% white.
+}
+
+void RgbSingleLED::turnOff() {
+    // Turn off the first LED to indicate "off" status.
+    //TODO
```
fiendie, line 25: See above

src/main.cpp
genkigenki: I started in the state machine, but its quite a lot of code duplication, as e.g. 
                case kBrew:
                case kShotTimerAfterBrew:
                case kBrewDetectionTrailing:
                case kBackflush:  
all have the same behaviour and the code below these the four cases would have to be repeated in four different places (in each "case" of the main state machine)

On top for each case there should be:
- The "if FEATURE.." check
- The "LEDupdate.check()"

Readability is also better, its easy to see what the behaviour is, and all LED code is in the loopLED() call.

Also it would not fit into the main state machine perfectly, as i want to implement a "cup illumination" for a longer time with smooth fade out. That wouldnt easily be possible in one location of the state machine, but rather a check before the loopLED switch case.

src/main.cpp
genkigenki: removed

src/hardware/RGBLEDstrip.cpp
```diff
@@ -0,0 +1,107 @@
+/**
+ * @file RGBLEDstrip.cpp
+ * 
+ * @brief A strip with adressable LEDs connected to a GPIO pin
+ */
+
+#include "RGBLEDstrip.h"
+#include "chipsets.h"
+#include "Logger.h"
```
fiendie, line 9: Include is never used. And those classes shouldn't have a dependency for the logger anyway.

src/hardware/RgbMultipleLED.cpp
```diff
@@ -0,0 +1,42 @@
+/**
+ * @file RgbMultipleLED.cpp
+ * 
+ * @brief Subgroup of adressable LEDs on the strip (all LEDs which are not the status or brew LED).
+ * 
+ * @param ledStrip The constructor saves the reference to the RGBLED object.
+ */
+
+#include "RgbMultipleLED.h"
+#include "Logger.h"
```
fiendie, line 10: See above.

src/main.cpp
genkigenki: Tried to reduce the switch-case to the maximum, still it seems cleaner to keep it in the same location (would need check if an error state is set to not overwrite two different colors with each other each loop().

PR #473, opened 2024-03-05
Adressable LEDs (like WS2811) in HAL

genkigenki:
Added adressable LEDs for the HAL.
Status / brew LEDs are automatically assigned first, remainder is assigned to context-LEDs.
All based on LED class: One RGBLEDstrip class provides access to WS2811/WS2812/... Additional classes for single-LEDs and multiple-LEDs. Cleanup of LED code in main.cpp

Detail:
LoopLED will be called every 80ms for relatively smooth LED animations.
Number of LEDs and chip type are configurable.

PID-on (not standby) -> low brightness white
Perfect temp -> dark green
Brew/Steam/Backflush -> white
Initialize, cold start Heat up -> Colors
Error message -> Red heartbeat (Water empty, sensor error, etc)
Steam-heat-up -> dark yellow

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -24,6 +24,7 @@ lib_deps =
     git+https://github.com/esphome/AsyncTCP#dc64fed
     git+https://github.com/esphome/ESPAsyncWebServer#f2a65ff
     git+https://github.com/tzapu/WiFiManager#71937d1
+    fastled/FastLED @ 3.6.0
 extra_scripts = pre:auto_firmware_version.py
 
 [env:esp32_usb]
--- a/src/brewHandler.h
+++ b/src/brewHandler.h
@@ -125,6 +125,8 @@ void checkbrewswitch() {
                 break;
 
             case kBrewSwitchFlushOff:
+                // Set white LED on, keep on even after manual brew and fade out
+                brewLEDon = BREW_FINISHED_LEDON_DURATION;
                 // Brew switch got released - stop flushing
                 if (currBrewSwitchStateMomentary == LOW && currStateBrewSwitch == LOW) {
                     brewSwitchState = kBrewSwitchReset;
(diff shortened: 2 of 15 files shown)


genkigenki:
[fiendie](https://github.com/fiendie) [Mar 5, 2024](https://github.com/rancilio-pid/clevercoffee/pull/472#discussion_r1512344638)

I suspect you're trying to get around having to write if (FEATURE_CONTEXT_LED) in every case-statement but duplicating the whole state machine seems infinitely worse.

Contributor Author
@[genkigenki](https://github.com/genkigenki) genkigenki [Mar 5, 2024](https://github.com/rancilio-pid/clevercoffee/pull/472#discussion_r1513523651) •

I started in the state machine, but its quite a lot of code duplication, as e.g.
case kBrew:
case kShotTimerAfterBrew:
case kBrewDetectionTrailing:
case kBackflush:
all have the same behaviour and the single code snippet below these the four cases would have to be repeated in four different places (in each "case" of the main state machine)

On top for each case there should be:

    The "if FEATURE.." check
    The "LEDupdate.check()"

Readability is also better, its easy to see the behaviour.
All LED update code is in the loopLED() call.

In the end, it would not fit into the main state machine perfectly, as i want to implement a "cup illumination" for a longer time with smooth fade out. That wouldnt easily be possible in each case of the main state machine, but rather a check before the loopLED switch case.

Contributor Author
@[genkigenki](https://github.com/genkigenki) genkigenki [Mar 5, 2024](https://github.com/rancilio-pid/clevercoffee/pull/472#discussion_r1513561356) •

Reduced the switch-case considerably, still its cleaner to keep it in the same location (would need check if an error state is set to not overwrite two different colors with each other each loop() in the loopLED and loopPID.

genkigenki:
Used "sync" function of Github for latest commit on master, it was screwing up the original branch. 
Created a new branch, removed the old one. This removed the PR as well, so had to redo it.
Open comment preserved above.

fiendie:
We're moving in a different direction with the LED concept. Thanks for your work!

PR #475, opened 2024-03-09
Add Formatting Checks and Tools

simonspa:
This PR adds automatic code -formatting to clevercoffee in different ways:

* As a `platformio` target: now, `pio run -t check-format` will check code formatting, `pio run -t format` will apply formatting
* As [pre-commit](https://pre-commit.com/) action. Pre-commit needs to be installed once via `pre-commit install` and will then automatically check formatting before every commit, and offer correction.
* As Github action. This runs the `check-format` target of PlatformIO automatically so the pipeline reflects the current formatting state. This currently depends on [a bug in the GH action](https://github.com/karniv00l/platformio-run-action/pull/5) we use.

I have tried to follow the formatting described in the `CONTRIBUTING.md` file bu hand-crafting the corresponding `clang-format` rules, to be found in `.clang-format`. However, there are a few caveats

* Many of the rules have not been applied entirely consistently, so this PR contains tons of changes. This is not very nice since other pending PRs and branches will very likely conflict and will need rebasing.
* There are some styles which I was not able to replicate with `clang-format`, specifically indentation of regular code following preprocessor macros, specifically `#if`:

    ```cpp
    #if BEWSOMETHING
        regular_code();
    #endif
    ```

    will inevitably become

    ```cpp
    #if BEWSOMETHING
    regular_code();
    #endif
    ```

    Maybe there is some magic I have not been able to squeeze out of `clang-format`, but maybe not. Possible ways forward apart from finding magic sauce:
        * Decide that we are fine with *not* indenting within preprocessor `#if`s
        * Add `// clang-format off` brackets around every occurrence
        * Scrap the entire formatting attempt

For the legibility of this PR, I have not committed the actual formatting changes yet, maybe run the targets locally and we can discuss the diffs and what to adjust in the formatting settings. I will then add the changes and we can move forward to merge.

### DIFF
--- a/lib/Logger/Logger.cpp
+++ b/lib/Logger/Logger.cpp
@@ -3,15 +3,18 @@
 
 #include "Logger.h"
 
-Logger::Logger(const uint16_t port)
-    : port_(port), server_(port) {}
+Logger::Logger(const uint16_t port) :
+    port_(port), server_(port) {
+}
 
 Logger& Logger::getInstanceImpl(const uint16_t port) {
     static Logger instance{port};
     return instance;
 }
 
-void Logger::init(const uint16_t port) { getInstanceImpl(port); }
+void Logger::init(const uint16_t port) {
+    getInstanceImpl(port);
+}
 
 bool Logger::begin() {
     if (WiFi.status() == WL_CONNECTED) {
@@ -40,7 +43,9 @@ bool Logger::update() {
     return true;
 }
 
-uint16_t Logger::getPort() { return Logger::getInstance().getPort(); }
+uint16_t Logger::getPort() {
+    return Logger::getInstance().getPort();
+}
 
 void Logger::log(const Level level, const String& file, const __FlashStringHelper* function, uint32_t line, const char* logmsg) {
     char time[12];
@@ -119,12 +124,19 @@ void Logger::current_time(char* timestamp) {
 
 String Logger::get_level_identifier(Logger::Level lvl) {
     switch (lvl) {
-        case Level::TRACE: return "  TRACE";
-        case Level::DEBUG: return "  DEBUG";
-        case Level::INFO: return "   INFO";
-        case Level::WARNING: return "WARNING";
-        case Level::ERROR: return "  ERROR";
-        case Level::FATAL: return "  FATAL";
-        default: return " SILENT";
+        case Level::TRACE:
+            return "  TRACE";
+        case Level::DEBUG:
+            return "  DEBUG";
+        case Level::INFO:
+            return "   INFO";
+        case Level::WARNING:
+            return "WARNING";
+        case Level::ERROR:
+            return "  ERROR";
+        case Level::FATAL:
+            return "  FATAL";
+        default:
+            return " SILENT";
     }
 }
--- a/lib/Logger/Logger.cpp
+++ b/lib/Logger/Logger.cpp
@@ -1,21 +1,20 @@
-#include <utility>
 #include <WiFi.h>
+#include <utility>
 
 #include "Logger.h"
 
-Logger::Logger(const uint16_t port) : port_(port), server_(port) {}
+Logger::Logger(const uint16_t port)
+    : port_(port), server_(port) {}
 
 Logger& Logger::getInstanceImpl(const uint16_t port) {
-    static Logger instance{ port };
+    static Logger instance{port};
     return instance;
 }
 
-void Logger::init(const uint16_t port) {
-    getInstanceImpl(port);
-}
+void Logger::init(const uint16_t port) { getInstanceImpl(port); }
 
 bool Logger::begin() {
-    if(WiFi.status() == WL_CONNECTED) {
+    if (WiFi.status() == WL_CONNECTED) {
         Logger::getInstance().server_.begin();
     }
 
@@ -41,15 +40,9 @@ bool Logger::update() {
     return true;
 }
 
-uint16_t Logger::getPort() {
-    return Logger::getInstance().getPort();
-}
+uint16_t Logger::getPort() { return Logger::getInstance().getPort(); }
 
-void Logger::log(const Level level,
-                    const String& file,
-                    const __FlashStringHelper* function,
-                    uint32_t line,
-                    const char* logmsg) {
+void Logger::log(const Level level, const String& file, const __FlashStringHelper* function, uint32_t line, const char* logmsg) {
     char time[12];
     current_time(time);
 
@@ -57,7 +50,7 @@ void Logger::log(const Level level,
         client_.print(time);
         client_.print(get_level_identifier(level).c_str());
         client_.print(" ");
-        if(level < Level::DEBUG) {
+        if (level < Level::DEBUG) {
             client_.print(file.c_str());
             client_.print(":");
             client_.print(line);
@@ -72,7 +65,7 @@ void Logger::log(const Level level,
         Serial.print(time);
         Serial.print(get_level_identifier(level).c_str());
         Serial.print(" ");
-        if(level < Level::DEBUG) {
+        if (level < Level::DEBUG) {
             Serial.print(file.c_str());
             Serial.print(":");
             Serial.print(line);
@@ -85,17 +78,13 @@ void Logger::log(const Level level,
     }
 }
 
-void Logger::logf(const Level level,
-                     const String& file,
-                     const __FlashStringHelper* function,
-                     uint32_t line,
-                     const char* format, ...) {
+void Logger::logf(const Level level, const String& file, const __FlashStringHelper* function, uint32_t line, const char* format, ...) {
     // reimplement printf method so that we can take dynamic list of parameters as printf does
     // (we can't simply pass these on to existing printf methods it seems)
 
     va_list arg;
     va_start(arg, format);
-    char temp[64];  // allocate a temp buffer
+    char temp[64]; // allocate a temp buffer
     char* buffer = temp;
     size_t len = vsnprintf(temp, sizeof(temp), format, arg);
     va_end(arg);
@@ -122,20 +111,20 @@ void Logger::logf(const Level level,
 
 void Logger::current_time(char* timestamp) {
     time_t rawtime;
-    struct tm *timeinfo;
+    struct tm* timeinfo;
     time(&rawtime);
     timeinfo = localtime(&rawtime);
     snprintf(timestamp, 12, "[%02d:%02d:%02d] ", timeinfo->tm_hour, timeinfo->tm_min, timeinfo->tm_sec);
 }
 
-String Logger::get_level_identifier(Logger::Level lvl){
-    switch(lvl) {
-        case Level::TRACE:   return "  TRACE";
-        case Level::DEBUG:   return "  DEBUG";
-        case Level::INFO:    return "   INFO";
+String Logger::get_level_identifier(Logger::Level lvl) {
+    switch (lvl) {
+        case Level::TRACE: return "  TRACE";
+        case Level::DEBUG: return "  DEBUG";
+        case Level::INFO: return "   INFO";
         case Level::WARNING: return "WARNING";
-        case Level::ERROR:   return "  ERROR";
-        case Level::FATAL:   return "  FATAL";
-        default:             return " SILENT";
+        case Level::ERROR: return "  ERROR";
+        case Level::FATAL: return "  FATAL";
+        default: return " SILENT";
     }
 }
--- a/lib/Logger/Logger.h
+++ b/lib/Logger/Logger.h
@@ -20,7 +20,9 @@ class Logger {
          * @brief Return singleton instance of the logger
          * @return Logger instance
          */
-        static Logger& getInstance() { return getInstanceImpl(); }
+        static Logger& getInstance() {
+            return getInstanceImpl();
+        }
 
         /**
          * @brief Initialize the singleton logger instance
@@ -82,8 +84,12 @@ class Logger {
          */
         void logf(const Level level, const String& file, const __FlashStringHelper* function, uint32_t line, const char* format, ...);
 
-        static void setLevel(Level level) { getInstance().level_ = level; }
-        static Level getCurrentLevel() { return getInstance().level_; }
+        static void setLevel(Level level) {
+            getInstance().level_ = level;
+        }
+        static Level getCurrentLevel() {
+            return getInstance().level_;
+        }
 
     private:
         static Logger& getInstanceImpl(const uint16_t port = 23);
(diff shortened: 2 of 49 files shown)


simonspa:
Voilà, the upstream GitHub Action bug has been merged, so now the action here fails because of the actual formatting. :)

Have a look and let me know so we can update the code formatting.

lib/Logger/Logger.cpp
```diff
@@ -45,19 +47,16 @@ uint16_t Logger::getPort() {
     return Logger::getInstance().getPort();
 }
 
-void Logger::log(const Level level,
-                    const String& file,
-                    const __FlashStringHelper* function,
-                    uint32_t line,
-                    const char* logmsg) {
+void Logger::log(const Level level, const String& file, const __FlashStringHelper* function, uint32_t line,
+                 const char* logmsg) {
```
LoQue90, line 51: This could be one line or should be split by "," like before.
i do prefer the split, like with our "editableVars".
if we go with one line, we have to find a way with the very long mashine state stuff.

lib/Logger/Logger.cpp
```diff
@@ -85,17 +84,14 @@ void Logger::log(const Level level,
...
-                     const char* format, ...) {
+void Logger::logf(const Level level, const String& file, const __FlashStringHelper* function, uint32_t line,
+                  const char* format, ...) {
     // reimplement printf method so that we can take dynamic list of parameters as printf does
     // (we can't simply pass these on to existing printf methods it seems)
 
     va_list arg;
     va_start(arg, format);
-    char temp[64];  // allocate a temp buffer
+    char temp[64]; // allocate a temp buffer
```
LoQue90, line 94: my ocd realy like the look of aligned comments.
maybe we go for a fixed spacing befor a comment?

lib/Logger/Logger.cpp
simonspa: Split is nicer, I agree. Added via

```
AlignAfterOpenBracket: Align
BinPackParameters: false
```

lib/Logger/Logger.cpp
simonspa: Phew, that's a tough one. `clang-format` aligned neighboring comment lines like this:

```cpp
 // Includes
#include "display/bitmaps.h" // user icons for display
#include "languages.h"       // for language translation
```

...but not when there is a single comment only. Adding a fixed offset might be possible but will likely break this alignment - and will look different depending on the context. Imagine a fixed 4-space offset:

```cpp
    char temp[64];        // Comment looks okay here
    char* buffer = temp;

    char tomp[64];        // Comment looks weird here because of random indent
    size_t len = vsnprintf(temp, sizeof(temp), format, arg);
```

simonspa: I think I found something in this direction, setting


```
AlignTrailingComments:
  Kind: Always
  OverEmptyLines: 2
```

LoQue90: Awesome!

lib/Logger/Logger.cpp
```diff
@@ -41,23 +40,17 @@ bool Logger::update() {
     return true;
 }
 
-uint16_t Logger::getPort() {
-    return Logger::getInstance().getPort();
-}
```
LoQue90, line 46: Before clang format it was right, or?

lib/Logger/Logger.cpp
```diff
@@ -1,21 +1,20 @@
-#include <utility>
 #include <WiFi.h>
+#include <utility>
 
 #include "Logger.h"
 
-Logger::Logger(const uint16_t port) : port_(port), server_(port) {}
+Logger::Logger(const uint16_t port)
+    : port_(port), server_(port) {}
```
LoQue90, line 7: New line after the ":" not before?

lib/Logger/Logger.cpp
```diff
@@ -85,19 +78,15 @@ void Logger::log(const Level level,
...
     // reimplement printf method so that we can take dynamic list of parameters as printf does
     // (we can't simply pass these on to existing printf methods it seems)
 
     va_list arg;
     va_start(arg, format);
-    char temp[64];  // allocate a temp buffer
+    char temp[64]; // allocate a temp buffer
     char* buffer = temp;
-    size_t len = vsnprintf(temp, sizeof(temp), format, arg);
+    size_t len   = vsnprintf(temp, sizeof(temp), format, arg);
```
LoQue90, line 89: That's one whitespace too much.

lib/Logger/Logger.h
```diff
@@ -7,22 +7,20 @@
...
+            TRACE   = 0,
+            DEBUG   = 1,
+            INFO    = 2,
             WARNING = 3,
-            ERROR = 4,
-            FATAL = 5,
-            SILENT = 6,
+            ERROR   = 4,
+            FATAL   = 5,
+            SILENT  = 6,
```
LoQue90, line 16: looks like some kinde of alignment between lines is going on here.

lib/Logger/Logger.h
```diff
@@ -7,22 +7,20 @@
...
         };
 
         /**
          * @brief Return singleton instance of the logger
          * @return Logger instance
          */
-        static Logger& getInstance() {
-            return getInstanceImpl();
-        }
+        static Logger& getInstance() { return getInstanceImpl(); }
```
LoQue90, line 23: as before, was fine before formatting

lib/Logger/Logger.h
```diff
@@ -31,7 +29,7 @@ class Logger {
          */
         static void init(const uint16_t port);
 
-        Logger(Logger const&) = delete;
+        Logger(Logger const&)         = delete;
```
LoQue90, line 32: toooo much whitespace

lib/Logger/Logger.h
```diff
@@ -68,11 +66,7 @@ class Logger {
          * @param line The line number of the log message
          * @param logmsg Log message to be sent as payload
          */
-        void log(const Level level,
-                 const String& file,
-                 const __FlashStringHelper* function,
-                 uint32_t line,
-                 const char* logmsg);
+        void log(const Level level, const String& file, const __FlashStringHelper* function, uint32_t line, const char* logmsg);
```
LoQue90, line 69: i thought we decided to keep the version with new lines after ","

src/brewHandler.h
```diff
@@ -10,67 +10,67 @@
...
+float weightBrew                   = 0;                        // weight value of brew
+float scaleDelayValue              = 2.5;                      // value in gramm that takes still flows onto the scale after brew is stopped
+bool scaleFailure                  = false;
+const unsigned long intervalWeight = 200;                      // weight scale
+unsigned long previousMillisScale;                             // initialisation at the end of init()
+HX711_ADC LoadCell(PIN_HXDAT, PIN_HXCLK);
+
+    #if SCALE_TYPE == 0
+HX711_ADC LoadCell2(PIN_HXDAT2, PIN_HXCLK);
+    #endif
```
LoQue90, line 73: okay look like if we decide to do line alignment for comments, this also will be the case with stuff likes this

src/display/displayTemplateScale.h
```diff
@@ -76,11 +74,11 @@ void printScreen() {
...
-        #if (FEATURE_PRESSURESENSOR == 1)
-            u8g2.setCursor(32, 46);
-            u8g2.print("P: ");
-            u8g2.print(inputPressure, 1);
-        #endif
+#if (FEATURE_PRESSURESENSOR == 1)
+        u8g2.setCursor(32, 46);
+        u8g2.print("P: ");
+        u8g2.print(inputPressure, 1);
+#endif
```
LoQue90, line 81: this lost the intendation after void printScreen. But that ist the problem with the makros you said?

src/hardware/IOSwitch.cpp
```diff
@@ -68,14 +67,8 @@ bool IOSwitch::longPressDetected() {
     return false;
 }
 
-GPIOPin& IOSwitch::getGPIOInstance() {
-    return gpio;
-}
+GPIOPin& IOSwitch::getGPIOInstance() { return gpio; }
```
LoQue90, line 70: as before

src/hardware/StandardLED.cpp
```diff
@@ -1,21 +1,18 @@
...
+    : gpio(gpioInstance) {}
 
-void StandardLED::turnOn() {
-    gpio.write(HIGH);
-}
+void StandardLED::turnOn() { gpio.write(HIGH); }
 
-void StandardLED::turnOff() {
-    gpio.write(LOW);
-}
```
LoQue90, line 18: this change should also not happen

src/languages.h
```diff
@@ -9,89 +9,89 @@
...
-#if TOF == 1
-static const char *langstring_waterempty = "Wasser leer";
-#endif
-
-static const char *langstring_wifirecon = "Wifi reconnect:";
-static const char *langstring_connectwifi1 = "1: Verbinde WLAN:";
-static const char *langstring_nowifi[] = {"Kein ", "WLAN"};
-
-static const char *langstring_error_tsensor[] = {"Fehler, Temp: ", "Temp.-Sensor ueberpruefen!"};
+    #if (DISPLAYTEMPLATE <= 4)
```
LoQue90, line 12: funny, it put the intendation fot the #if behainde an #if, but can´t set it right for the content :D

lib/Logger/Logger.h
fiendie: If it's done automatically and somewhat consistently I don't really have an issue with it. Hand-formatting that stuff is incredibly tedious, though.

lib/Logger/Logger.h
fiendie: Yeah, I'm not a fan. 
Return statements belong on a separate line.

src/brewHandler.h
fiendie: OK, this is weird. 
The "canonical" way to align preprocessor macros is no indentation at all. I thought clang-tidy would refuse to do anything else :)

lib/Logger/Logger.cpp
simonspa:
```
BreakConstructorInitializers: AfterColon
```

check.

lib/Logger/Logger.cpp
simonspa:
```
AllowShortFunctionsOnASingleLine: false
```

check.

lib/Logger/Logger.cpp
simonspa: It depends. I enabled alignment of blocks of `=`. If this is not wanted I can switch it off again:

```
AlignConsecutiveAssignments: Consecutive
```

lib/Logger/Logger.h
simonspa: If you have enabled `pre-commit` it is automatic. `pio run -t format` also fixes this automatically.

lib/Logger/Logger.h
simonspa: Aye, see above, switched off.

lib/Logger/Logger.h
simonspa: See above, `=` alignment currently active.

lib/Logger/Logger.h
simonspa: That only happens when the line becomes too long. I think we can also switch it *always* on if you wish, but I'm not sure that will help. @fiendie wanted longer lines, if we break at every comma, we will hardly surpass the 20 characters per line :D

src/brewHandler.h
simonspa: I tried very hard to mimic your current style of using `#if` macros. I don't like it either and would rather fully left-align macros. If they are used as code, we should just change them for proper C++ code

```cpp
if (SCALE_TYPE == 0) {
    // ...
}
```

then it will also be aligned normally. I would do this in a separate step though. Opinions?

src/display/displayTemplateScale.h
simonspa: Exactly. If we change the `#if` macros for proper `if(){}` statements this will be fixed. See above.

src/hardware/IOSwitch.cpp
simonspa: Check, fixed.

src/hardware/StandardLED.cpp
simonspa: Jep, was also the "allow short functions on single line" which I now switched off.

src/languages.h
simonspa: Yes, because the `#if` are prepocessor macros and have nothing to do with the actual code. I find the general use of preprocessor stuff for `if(){}` problematic, but we can easily change that - and I understand you are working to get rid of them anyway...

src/brewHandler.h
LoQue90: fully left aligne for #if, and we get them out of the code soon

lib/Logger/Logger.cpp
LoQue90: as desided in the next comment, leave it turned on :)

src/brewHandler.h
simonspa: Done, pushed.

lib/Logger/Logger.h
kjyv: Do we really need this alignment? I think it's pretty unusual and not really helpful

src/display/displayTemplateStandard.h
```diff
@@ -75,7 +72,8 @@ void printScreen()
         u8g2.print("|");
 
         if (bPID.GetKi() != 0) {
-            u8g2.print(bPID.GetKp() / bPID.GetKi(), 0);;
+            u8g2.print(bPID.GetKp() / bPID.GetKi(), 0);
+            ;
```
kjyv, line 76: not really because of this PR but this extra ; can probably be deleted ;)

lib/Logger/Logger.h
simonspa: Indeed a bit unusual - but we also have unusually long lists of preprocessors directives, global variables and all. I found it helpful in reading - but can of course drop it if you wish.

LoQue90: Then I would take it out!

simonspa: Done.

src/display/displayTemplateStandard.h
simonspa: Removed.

LoQue90 (approved):

lib/Logger/Logger.h
kjyv: I also think it's nicer to only break when the line becomes too long

PR #480, opened 2024-03-11
refactor brew, steam, brewdetection and optocoupler

LoQue90:
this PR will remove the software brewdetection.
also steam over optocoupler for quickmill.
on default there will be no brewdetection anymore.
Since we change the behavior of the PID in total and we do not recommend heating during brew, there is no need for SW brewdetection.
On top there is a lot of code, that specialy is waiting because of SW brewdetection.
If someone want BREW PID, having some sort of hardware brewdetection (switch or optocoupler), is needed.
If we add brew detection with a reed sensor on the valve, the impact of adding a hardware brew detection is reduced.

During this PR i also want to rename Brewdection flags and variables, the fit our naming sheme and get this all a bit nicer.
Also there is maybe a bug that our main state mashine dosen´t travel back from ksteam to kpidnormal, when no brewdetection is enabled. This has to be fixed with this pr, or it isn´t possible to remove sw bredetection.

TODO:
Check code in total if anywhere some sw brewdecetion code is left
refactor brew, steam, brewdetection where neede, part of shotimer
rename variables
real world testing.

ANY help is highly welcome, feel free to comment and discuss. Or make PRs on my branch.

### DIFF
--- a/src/embeddedWebserver.h
+++ b/src/embeddedWebserver.h
@@ -60,7 +60,7 @@ void serverSetup();
 void setEepromWriteFcn(int (*fcnPtr)(void));
 
 // editable vars are specified in main.cpp
-#define EDITABLE_VARS_LEN 33
+#define EDITABLE_VARS_LEN 32
 extern std::map<String, editable_t> editableVars;
 
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -653,11 +653,15 @@ void brewDetection() {
     if (brewDetectionMode == 0) 
     return;  
 
-    if (brewDetectionMode == 1) {  // Hardware Switch
-        if (millis() - timeBrewDetection > brewtimesoftware * 1000 && isBrewDetected == 1) {
-            isBrewDetected = 0;  // rearm brewDetection
+     if (brewDetectionMode == 1) {  // Hardware Switch
+        if (currBrewState > kBrewIdle && brewDetected == 0) {
+            LOG(DEBUG, "HW Brew detected");
+            timeBrewDetection = millis();
+            isBrewDetected = 1;
+            brewDetected = 1;
         }
     }
+
     else if (brewDetectionMode == 2) { // optocoupler
         // timeBrewed counter
         if ((digitalRead(PIN_BREWSWITCH) == optocouplerOn) && brewDetected == 1) {
@@ -677,16 +681,6 @@ void brewDetection() {
         }
     }
 
-    // Activate brew detection
-
- if (brewDetectionMode == 1) {  // Hardware Switch
-        if (currBrewState > kBrewIdle && brewDetected == 0) {
-            LOG(DEBUG, "HW Brew detected");
-            timeBrewDetection = millis();
-            isBrewDetected = 1;
-            brewDetected = 1;
-        }
-    }
     else if (brewDetectionMode == 2) {  // optocoupler
         switch (machine) {
             case QuickMill:
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -475,63 +475,6 @@ void testEmergencyStop() {
 }
 
 /**
- * @brief FIR moving average filter for software brew detection
- */
-void calculateTemperatureMovingAverage() {
-    const int numValues = 15;                      // moving average filter length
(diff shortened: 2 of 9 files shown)


src/storage.h
```diff
@@ -93,18 +93,18 @@ typedef struct __attribute__((packed)) {
     uint8_t freeToUse8[2];
     double pidTvBd;
     uint8_t freeToUse9[2];
-    double brewSwTimeSec;
+    double freeToUse10;
```
kjyv, line 96: You can also rename some of previous ones too to get continuous counting

src/steamHandler.h
```diff
@@ -20,20 +20,6 @@ void checkSteamSwitch() {
             if (steamSwitchReading == LOW && !steamFirstON) {
                 steamON = 0;
             }
-
```
kjyv, line 23: Is this really related to brew detection? It looks like this is steam mode handling for quickmill machines

src/main.cpp
```diff
@@ -705,131 +616,6 @@ char *number2string(unsigned int in) {
     return number2string_uint;
 }
 
-
```
kjyv, line 708: this handles all brew detection, not just sw bd

src/main.cpp
```diff
@@ -853,32 +639,6 @@ void setEmergencyStopTemp() {
     }
 }
 
-void initSteamQM() {
```
kjyv, line 856: This is also steam mode for QM, not brew detection and should probably stay

src/main.cpp
```diff
@@ -1821,7 +1510,7 @@ void setup() {
         .type = kDouble,
         .section = sBDSection,
         .position = 20,
-        .show = [] { return true && FEATURE_BREWDETECTION == 1 && useBDPID; },
```
kjyv, line 1824: You'll probably want to remove all of these vars as their show condition is only true in case the BD feature is enabled and the flag useBDPID is true. That will never be the case when you remove the feature

src/main.cpp
```diff
@@ -1136,52 +883,11 @@ void handleMachineState() {
...
         case kShotTimerAfterBrew:
-            brewDetection();
 
             if (millis() - lastBrewTimeMillis > SHOTTIMERDISPLAYDELAY) {
                 LOGF(INFO, "Shot time: %4.1f s", lastBrewTime / 1000);
-                machineState = kBrewDetectionTrailing;
-                lastBrewTime = 0;
-            }
-
-            if (steamON == 1) {
```
kjyv, line 1147: Why remove all of this?

src/embeddedWebserver.h
```diff
@@ -60,7 +60,7 @@ void serverSetup();
 void setEepromWriteFcn(int (*fcnPtr)(void));
 
 // editable vars are specified in main.cpp
-#define EDITABLE_VARS_LEN 33
+#define EDITABLE_VARS_LEN 32
```
kjyv, line 63: Shouldn't this be 31 (two vars removed)?

kjyv:
It seems you're removing brew detection completely and also steam mode for quick mill - also the hardware based. Did I miss something or is this likely not what we want?

LoQue90:
@kjyv didn't had time to work on this, so right now it's a mess.

We did discussed the quickmill steam stuff, because it was somehow not working anymore. In the end we did find a bit confusing how this works. One idea for quickmill thermoblock was to remove main power from steam switch and connect it to the esp, but also keep the pulsor board ad trigger that with a ssr directly from the steam switch.
That's why I removed also all the quickmill steam stuff.
But I'm total open for any other solutions.
After removing the quickmill stuff from the optocoupler, nothing realy was left from brewdetection. So I started to removed this in total.
My idea is: with Feature_Brewswitch we start reading the switch, with Brewcontrol_type 0 we only bring the state mashine into kbrew, start the shottimer and trigger Brew pid if needed.
If Brewcontrol_type is > 0 we add brew() or brewscale () to control brewing.
I put that idea on discord, but didn't get a lot responses to this, I'm open for input!
Gonna rename this into refactor for brewdetection, brewswitch, steamswitch, optocoupler aaaaaannnnd more :D

PR #484, opened 2024-03-25
Fixed the scale readout with timed brewmode selected

jsve-dev:
Fixed shottimerscale(); never getting called with appropriate settings enabled.

Ps. this is my first pull request. I have about -18 idea what im doing... Hope im helping 😀

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1855,7 +1855,7 @@ void looppid() {
 
     handleMachineState();
 
-#if (FEATURE_SCALE == 1 && BREWCONTROL_TYPE == 0) // SHOTTIMER with scale
+#if (FEATURE_SCALE == 1 && BREWCONTROL_TYPE < 2) // SHOTTIMER with scale
     shottimerscale();
 #endif
 


LoQue90:
I'm very sorry, but this pr is doing a lot of stuff, but nothing that will actually change something helpful. 
What do you want to achieve?

LoQue90:
Okay if I look only at the total diff, it's okay.
But it should only be one commit.
This shottimer scale call should only be done for people which are using only the pid and the scale.
If you are using the scale and controlling the brew, the scale shottimer is called differently.

fiendie:
> Okay if I look only at the total diff, it's okay. But it should only be one commit. This shottimer scale call should only be done for people which are using only the pid and the scale. If you are using the scale and controlling the brew, the scale shottimer is called differently.

I've been chatting with him about this yesterday. I'll help squash the commits :)

jsve-dev:
Cleaned up the mess... Now it just the useful change.
Thanks @fiendie

LoQue90:
Can you tell my why do you want that change?
What is your setup?

jsve-dev:
In the case of full expansion with valve control and scale. But still brew by time with continuous readout of scale output.
Without calling shottimerscale() the scale turns off during a brew and the readout shows 0.0g.

I of course had not thought about the pid-only usecase, but changing the line to BREWCONTROL_TYPE <= 1 should cover both instances

LoQue90:
Should be < 2 to not make weird stuff if someone uses brew by scale.

jsve-dev:
Changed to <2

LoQue90:
@FabianSperrle but you also have brewcontrol 2 set?
This one should fix the scale readout for onlypidscale.
If I'm correct.

LoQue90 (approved):

PR #486, opened 2024-03-30
Draft: Drop ColdStart and AtSetPoint States

simonspa:
This PR drops the `ColdStart` state. The only difference with respect to `PidNormal` was that it made use of *PonM* in case it was activated. `PidNormal` always ignored this setting.

This PR moves *PonM* to the `PidNormal` state.

The `AtSetPoint` state was only entered from the `ColdStart` state and did not serve any purpose, hence it is also removed.

Without the `ColdStart` state, the condition for displaying the heating logo had to be changed. It is now displayed whenever in the `PidNormal` state *and* when the temperature is at least 5degC below the setpoint.

### DIFF
--- a/src/display/displayCommon.h
+++ b/src/display/displayCommon.h
@@ -310,7 +310,8 @@ void displayShottimer(void) {
  * @brief display heating logo
  */
 void displayMachineState() {
-    if (FEATURE_HEATINGLOGO > 0 && (machineState == kInit || machineState == kColdStart) && brewSwitchState != kBrewSwitchFlushOff) {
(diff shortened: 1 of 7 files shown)


simonspa:
Current `master`, no *PonM* in `PidNormal` state:

image

Current this branch, *PonM* active in `PidNormal` state (default start values):

cycle-ponm

fiendie:
I always thought the idea of a separate cold start phase was a bit overrated because the group needs time to heat up.

fiendie (approved):

kjyv (approved):
nice

PR #488, opened 2024-04-12
Drop Cooldown State

simonspa:
This MR does two things:

* It drops the `kCooldown` state since that just meant PID off (which is anyway the case because we're off the target value) and seeting a different emergency temperature
* It fixes a single emergency temperature to 145degC instead of swapping between 120 and 145. This is more like the original Silvia hardware where a single thermostat is used for the emergency shutoff (still in use)

### DIFF
--- a/src/display/displayTemplateScale.h
+++ b/src/display/displayTemplateScale.h
@@ -5,6 +5,7 @@
  *
  */
 
+#pragma once
 #include "displayCommon.h"
 
 /**
--- a/src/display/displayTemplateStandard.h
+++ b/src/display/displayTemplateStandard.h
@@ -5,6 +5,7 @@
  *
  */
 
+#pragma once
 #include "displayCommon.h"
 
 /**
--- a/src/display/displayTemplateUpright.h
+++ b/src/display/displayTemplateUpright.h
@@ -19,7 +19,7 @@ void printScreen() {
 
     // If no specific machine state was printed, print default:
     if (((machineState == kPidNormal || machineState == kBrewDetectionTrailing) || ((machineState == kBrew || machineState == kShotTimerAfterBrew) && FEATURE_SHOTTIMER == 0) || // shottimer == 0, auch Bezug anzeigen
-         machineState == kCoolDown || (machineState == kPidNormal && (setpoint - temperature) > 5. && FEATURE_HEATINGLOGO == 0) || ((machineState == kPidDisabled) && FEATURE_PIDOFF_LOGO == 0)) &&
+         (machineState == kPidNormal && (setpoint - temperature) > 5. && FEATURE_HEATINGLOGO == 0) || ((machineState == kPidDisabled) && FEATURE_PIDOFF_LOGO == 0)) &&
         (brewSwitchState != kBrewSwitchFlushOff)) {
         if (!sensorError) {
             u8g2.clearBuffer();
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -276,7 +276,7 @@ SysPara<float> sysParaScaleKnownWeight(&scaleKnownWeight, 0, 2000, STO_ITEM_SCAL
 
 // Other variables
 boolean emergencyStop = false;                // Emergency stop if temperature is too high
-double EmergencyStopTemp = 120;               // Temp EmergencyStopTemp
+const double EmergencyStopTemp = 145;         // Temp EmergencyStopTemp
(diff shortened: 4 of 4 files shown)


simonspa:
...seems like I'm much better in *deleting code* than in *writing code* :rofl:

simonspa:
Nota bene: for Silvia, the `kCooldown` state had exactly the same PID values as for `kPidNormal`, so virtually no difference deleting it. For Quickmill thermoblock machines, `ki` and `kd` were set to 0, so here it would be interesting to get a comparison. Anyone has a thermoblock around? :)

simonspa:
Corresponding UML diagram (omitting error states) after this PR:

state_machine_nocooldown

kjyv:
Maybe you can find out from the git history who added the kCooldown state originally so we understand the idea behind it? It might not be doing anything (anymore) for Silvia but setting the params to 0 for thermoblocks sounds like this would really cool down the thermoblock for some time - not sure why.

simonspa:
git show -p 6be4086fc3321fe3e11620a9cc9de601c4478eb0

>  Abkühlmodus "chill-mode after steam": Neuer machinestate 45 für die Abkühlphase nach dem Dampfmodus. In diesem maschinestate kann mit dem Brühschalter ein cooling flush gemacht werden oder mit dem Dampfschalter auf steam geschaltet werden.
    Wenn bis auf Solltemperatur + 2 °C abgekühlt ist, wird in den idle-Betrieb (machinestate 20) gewechselt.
    Für die QuickMill werden im machinestate 45 die PID-Werte auf Dampfeinstellungen belassen, für alle anderen Maschinen wird auf die Brüh-PID-Werte umgeschaltet.

kjyv:
Unexpectedly detailed commit message, nice :) 
I'm not sure why it would make sense to keep steam params for quickmill for cool down but I also don't know how heating is done differently for those machines. I can see that we have a cooling flush detection and detection when the steam switch is off for those machines. But all of that should still work the same when brewDetection() is called from other states than kCooldown

kjyv:
This part in the commit might also be important:
```
    if (Brewdetection == 2 || Brewdetection == 3) 
      {
        /*
          Bei QuickMill Dampferkennung nur ueber Bezugsschalter moeglich, durch Aufruf von 
          brewdetection() kann neuer Dampfbezug erkannt werden
          */ 
        brewdetection();
      }
```

LoQue90:
We where talking about getting all that quickmill stuff out of the code.
This is detects steam on a thermoblock while counting the pulses of the pump. A way to controll steam, but don't have the pump controlled from the esp.
In Discord we found a solution, where we connected the steam switch with the esp and trigger the og pump pulsing pcb with a Relais.
In long term we would like to do the pump pulsing completely in software.

kjyv:
If I understand correctly, there is no problem if this causes any issues (unlikely but not 100% sure) with quick mill machines as all support is going to be dropped anyway?

LoQue90:
We would like to go for a general generic support for boiler and thermoblock mashines. So anything that is specific for quickmill has either to go or should be rewrite to be generic.
In the end there will be no support anymore for this pump pulse counting stuff, as at the moment, we don't see a benefit from controlling the steam temp of thermoblocks, but not controlling also the pump.
So there should be no problem in removing cool down I guess.

simonspa:
Rebased on current `master`

kjyv:
What do we need here in order to merge? The comments I made were only intended to help fuel the discussion, but I have no strong opinion about Quickmill heritage. And if the plan is to remove the current code for it anyway, then I'd say we can go ahead, wdyt @fiendie?

LoQue90:
I wasn't at home the last 3 weeks, so no chance in testing this.
On top we don't know yet if the cooldown is in any way needed for thermoblock.

fiendie:
> What do we need here in order to merge? The comments I made were only intended to help fuel the discussion, but I have no strong opinion about Quickmill heritage. And if the plan is to remove the current code for it anyway, then I'd say we can go ahead, wdyt @fiendie?

The new state was introduced with commit 6be4086fc3321fe3e11620a9cc9de601c4478eb0 in 2021. It's certainly not needed and I'm personally a bit skeptical about the efficacy of changing the PID settings for different phases especially when it comes to cooldown. 

I briefly tested the PR and I'd say we just merge it :)

fiendie (approved):

LoQue90:
Add, didn't know that you tested it, awesome!

PR #491, opened 2024-04-13
Timer Class

simonspa:
This PR contains a super-simple timer class which can be used in lieu of manual `millis()` checks and corresponding global variables. Right now it only takes a `void()` function pointer (no arguments, no return values) but if this turns out to be useful in removing clutter, this can be extended.

I used the water sensor as an example for to use it:

* Create a `Timer` object with the target function to call and the interval
* Call the timer objects `operator()` in every `loop()` iteration


fiendie:
While you're at it, would you mind moving the PeriodicTrigger to the newly created utils folder as well? 😄

simonspa:
They are indeed quite similar. Might simplify this in the future. Moved for now.

fiendie (approved):

PR #493, opened 2024-04-13
Improve Switch Classes

simonspa:
This makes a few changes to the switch classes:

* `Switch` now holds its own `Type` and `Mode`
* `IOSwitch` now holds its private `GPIOPin` and also constructs it

This simplifies the overall code, encapsulates GPIO pins where they belong and removes global pointer objects.


simonspa:
(cc) @fiendie

fiendie (approved):

PR #496, opened 2024-04-15
Fix Shottimer Scale and inclusion of BREWCONTROL_TYPE 2 in 1, with choice of brew-type via parameter in Website

TobiEdelberg:
brewHandler.h/userConfig_sample.h:
Removed the brew by weight as a separate type and included its functionality in brew by time for easier maintenance.

displayCommon.h:
Updated SHOTTIMER_TYPE 2 (With scale) to functionality already available at SHOTTIMER_TYPE 1

### DIFF
--- a/src/brewHandler.h
+++ b/src/brewHandler.h
@@ -229,7 +229,7 @@ void backflush() {
     }
 }
 
-#if (BREWCONTROL_TYPE == 1)
+#if (FEATURE_BREWCONTROL == 1)
 /**
  * @brief Time base brew mode
  */
@@ -252,7 +252,13 @@ void brew() {
         brewSwitchWasOff = true;
     }
 
-    totalBrewTime = (preinfusion * 1000) + (preinfusionPause * 1000) + (brewTime * 1000); // running every cycle, in case changes are done during brew
+    if (brewTime > 0) {
+        totalBrewTime = (preinfusion * 1000) + (preinfusionPause * 1000) + (brewTime * 1000); // running every cycle, in case changes are done during brew
+    }
+    else {
+        // Stop by time deactivated --> brewTime = 0
+        totalBrewTime = 0;
+    }
 
     // state machine for brew
     switch (currBrewState) {
@@ -311,12 +317,19 @@ void brew() {
 
             break;
 
-        case kWaitBrew: // waiting time brew
+        case kWaitBrew: // waiting time or weight brew
             lastBrewTime = timeBrewed;
 
-            if (timeBrewed > totalBrewTime) {
+            // stop brew if target-time is reached --> No stop if stop by time is deactivated via Parameter (0)
+            if ((timeBrewed > totalBrewTime) && ((brewTime > 0))) {
                 currBrewState = kBrewFinished;
             }
+#if (FEATURE_SCALE == 1)
+            // stop brew if target-weight is reached --> No stop if stop by weight is deactivated via Parameter (0)
+            else if (((FEATURE_SCALE == 1) && (weightBrew > weightSetpoint)) && (weightSetpoint > 0)) {
+                currBrewState = kBrewFinished;
+            }
+#endif
 
             break;
 
@@ -325,7 +338,6 @@ void brew() {
             valveRelay.off();
             pumpRelay.off();
             currBrewState = kWaitBrewOff;
-            timeBrewed = 0;
 
             break;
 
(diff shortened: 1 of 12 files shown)


LoQue90:
this breaks brew by time if no scale is connected

TobiEdelberg:
Merged with actual changes of Master today.

src/brewHandler.h
```diff
@@ -314,9 +335,17 @@ void brew() {
...
 
+            if (FEATURE_SCALE == 1) {
+                weightBrew = weight - weightPreBrew;
+            }
+
             if (timeBrewed > totalBrewTime) {
                 currBrewState = kBrewFinished;
             }
+            // stop brew if target-weight is reached before target-time.
+            else if ((FEATURE_SCALE == 1) && (weightBrew > (weightSetpoint - scaleDelayValue))) {
```
FabianSperrle, line 346: I think this condition should not check `FEATURE_SCALE` but `BREWCONTROL_TYPE == 2` -- otherwise, stopping the brew might be unexpected behaviour when a scale is connected but brewing is supposed to only stop by time. 

Similarly, I think that the lines above should be behind a BREWCONTROL_TYPE == 1 flag...

src/brewHandler.h
```diff
@@ -314,9 +335,17 @@ void brew() {
         case kWaitBrew: // waiting time brew
             lastBrewTime = timeBrewed;
 
+            if (FEATURE_SCALE == 1) {
+                weightBrew = weight - weightPreBrew;
+            }
+
             if (timeBrewed > totalBrewTime) {
```
FabianSperrle, line 342: conditional onl `BREWCONTROL_TYPE == 1`?

TobiEdelberg:
Check with condition FEATURE_SCALE is correct as in my PR there is no BREWCONTROL_TYPE 2 anymore available/needed. There is no unexpected behavior, because you set target weight and target time in the web-ui and whatever comes first will stop the brew. Priority can be set by the targets entered. 

I don’t know what you mean by 

„conditional onl BREWCONTROL_TYPE == 1?“

> Am 20.05.2024 um 22:19 schrieb Fabian Sperrle-Roth ***@***.***>:
> I think this condition should not check FEATURE_SCALE but BREWCONTROL_TYPE == 2 -- otherwise, stopping the brew might be unexpected behaviour when a scale is connected but brewing is supposed to only stop by time.

LoQue90:
What is with people that just want to see the weight, without stopping at the set weight?

TobiEdelberg:
Just increase the target weight to a higher value (e.g. 100g) and it will stop by time but you see the weight. You are still able to decide without rebulding the code what you want as a target in the web-UI and can change on the fly without a pc. Am 20.05.2024 um 23:49 schrieb Lorenz Uhlig ***@***.***>:﻿
What is with people that just want to see the weight, without stopping at the set weight?

LoQue90:
Okay, but then we should do this right:
Brewcontrol_type should be renamed to Feature_Brewcontrol
Feature_Brewcontrol is then like Onlypid
With Feature_Scale 1 will see the weight like right now with onlypidscale
And shottimertype 1 and 2 should also just be Feature_Shottimer and with Feature_Scale you will see the weight or not.

This would be all the stuff that should be done with this pr.
@FabianSperrle @fiendie @TobiEdelberg

TobiEdelberg:
Fine for me. 😉 Will have a look today or tomorrow in the evening. Am 21.05.2024 um 06:39 schrieb Lorenz Uhlig ***@***.***>:﻿
Okay, but then we should do this right:
Brewcontrol_type should be renamed to Feature_Brewcontrol
Feature_Brewcontrol is then like Onlypid
With Feature_Scale 1 will see the weight like right now with onlypidscale
And shottimertype 1 and 2 should also just be Feature_Shottimer and with Feature_Scale you will see the weight or not.
This would be all the stuff that should be done with this pr.
@FabianSperrle @fiendie @TobiEdelberg

FabianSperrle:
Setting large brew times or large target weights to "disable" brew by time or brew by weight, respectively, feels odd to me.
My intuition would have been to add a new `BREWCONTROL_TYPE  3` that stops by time or by weight (whatever comes first) to complement the existing features. Additionally, switching the `BREWCONTROL_TYPE` could be made available via the web UI and the menu. 

However, that's more complexity again, and maybe sometimes the simple solutions are good enough...

LoQue90:
I'm totally with you @FabianSperrle, I don't realy like the solution.
Adding a 3 option, setting in userconfig would be very easy.
To select this via web interface have to wait until we switched to a config file.
And while this is just a copy of existing code, there isn't much complexity added.

TobiEdelberg:
Possible to do so but as you describe correctly this would be the more complex solution and therefore I made a decision against it. Also there is no need to use high values just to prioritize one of the targets or just eliminate one completely. My usage:I added just the target weight I want to get and add two seconds on top of the max time I want to have a sort of timeout for my target weight. Also possible to use both targets with the correct values you want and whatever comes first wins. 😉 No disadvantage in my opinion and it is possible to fine tune with every shot.Am 21.05.2024 um 18:14 schrieb Fabian Sperrle-Roth ***@***.***>:﻿
Setting large brew times or large target weights to "disable" brew by time or brew by weight, respectively, feels odd to me.
My intuition would have been to add a new BREWCONTROL_TYPE  3 that stops by time or by weight (whatever comes first) to complement the existing features. Additionally, switching the BREWCONTROL_TYPE could be made available via the web UI and the menu.
However, that's more complexity again, and maybe sometimes the simple solutions are good enough...

src/main.cpp
FabianSperrle, line 1: There are still references to `BREWDETECTION_TYPE` left, this does not even compile.

FabianSperrle (changes requested):

src/main.cpp
TobiEdelberg: BREWDETECTION_TYPE was not changed. 

Discussion was to rename BREWCONTROL_TYPE to FEATURE_BREWCONTROL. 

Did you pull the whole branch and compile? 
Did you use the updated userConfig_sample.h?
What settings do you use in your userConfig?

LoQue90:
No, it is not just a renaming, taking our brewcontrol type in total is a bit more work.
But @FabianSperrle and I prefer to add a 3. Control type.

src/brewHandler.h
```diff
@@ -314,9 +322,17 @@ void brew() {
         case kWaitBrew: // waiting time brew
             lastBrewTime = timeBrewed;
 
+            if (FEATURE_SCALE == 1) {
+                weightBrew = weight - weightPreBrew;
+            }
+
```
FabianSperrle, line 328: Why does this have to be done here explicitly? Is it not handled by the general block above in line 250 already?

FabianSperrle:
```
[01:28:46]   DEBUG brewSwitchState = kBrewSwitchIdle; waiting for brew switch input
[01:28:47]   DEBUG brewSwitchState = kBrewSwitchBrew; brew switch short pressed - start Brew
[01:28:47]   DEBUG HW Brew detected
[01:28:47]    INFO Resetting standby timer to 30 minutes
[01:28:47]   DEBUG new machineState: PID Normal -> Brew
[01:28:47]   DEBUG new PID-Values: P=62.0  I=1.2  D=1116.0
[01:28:47]    INFO Brew started
[01:28:47]   DEBUG disabled PID, waiting for 0 seconds before enabling PID again
[01:28:55]    INFO Standby time remaining: 29 minutes
[01:28:57]   DEBUG Enabled PID again after delay
[01:29:00]    INFO Brew stopped
[01:29:00]   DEBUG new machineState: Brew -> Shot Timer After Brew
[01:29:00]   DEBUG new PID-Values: P=62.0  I=1.2  D=1116.0
[01:29:00]   DEBUG brewSwitchState = kBrewSwitchBrewAbort: brew switch short pressed - stop brew
[01:29:00]    INFO Brew stopped manually
[01:29:00]   DEBUG brewSwitchState = kBrewSwitchReset: brew switch released - go to kBrewSwitchIdle
[01:29:03]    INFO Shot time: 13.5 s
[01:29:03]   DEBUG new machineState: Shot Timer After Brew -> Brew Detection Trailing
[01:29:03]   DEBUG new PID-Values: P=62.0  I=1.2  D=1116.0
[01:29:12]   DEBUG new machineState: Brew Detection Trailing -> PID Normal
[01:29:12]   DEBUG new PID-Values: P=62.0  I=1.2  D=1116.0
```

Was die state machine da logged sieht in bisschen wild aus, ich habe den Brew nicht händisch gestoppt sondern das Gewicht war erreicht... Aber das hat nichts mit dem PR zu tun denke ich.

LoQue90 (dismissed):

src/brewHandler.h
```diff
@@ -55,15 +55,14 @@ boolean brewPIDDisabled = false; // is PID disabled for delay after brew has sta
...
+bool scaleFailure = false;
 boolean scaleCalibrationOn = 0;
 boolean scaleTareOn = 0;
 int shottimerCounter = 10;
 float calibrationValue = SCALE_CALIBRATION_FACTOR; // use calibration example to get value
-float weight = 0;                                  // value from HX711
-float weightPreBrew = 0;                           // value of scale before wrew started
-float weightBrew = 0;                              // weight value of brew
-float scaleDelayValue = 2.5;                       // value in gramm that takes still flows onto the scale after brew is stopped
-bool scaleFailure = false;
```
FabianSperrle, line 66: Why?

src/main.cpp
```diff
@@ -1304,11 +1301,11 @@ void setup() {
 
     editableVars["BREW_TIME"] = {.displayName = F("Brew Time (s)"),
                                  .hasHelpText = true,
-                                 .helpText = F("Stop brew after this time"),
+                                 .helpText = F("Stop brew after this time --> 0=Deactivated"),
```
FabianSperrle, line 1304: I don't like the --> notation, does not look very finished... 

How about "Stop brew after this time. Set to 0 to deactivate brew by time." ?

src/main.cpp
```diff
@@ -1328,18 +1325,18 @@ void setup() {
...
-                                        .show = [] { return true && BREWCONTROL_TYPE > 0; },
+                                        .show = [] { return true && FEATURE_BREWCONTROL == 1; },
                                         .minValue = PRE_INFUSION_TIME_MIN,
                                         .maxValue = PRE_INFUSION_TIME_MAX,
                                         .ptr = (void*)&preinfusion};
 
     editableVars["SCALE_WEIGHTSETPOINT"] = {.displayName = F("Brew weight setpoint (g)"),
                                             .hasHelpText = true,
-                                            .helpText = F("Brew until this weight has been measured."),
+                                            .helpText = F("Brew until this weight has been measured. --> 0=Deactivated"),
```
FabianSperrle, line 1337: See above.

FabianSperrle (approved):
Looks good to me, works on my machine.

I have minor comments on the wordings of the website tooltip, but that might just be my preference.

src/brewHandler.h
LoQue90: this comes from moving these vars out of the FEATURE_SCALE flag and back into it, could be done cleaner -> not touching them at all in the end.
We changed our minds a lot douring this pr

PR #503, opened 2024-04-25
TFT support with eSPI driver

aschoelzhorn:
Support for eSPI driver for TFT displays


fiendie:
Unfortunately this doesn't really fit in with our current hardware layout. Thank you for submitting anyway!

PR #505, opened 2024-05-08
Fix Emergency mode

theHolgi:
**Setup**: 
Machine: Rancilio Silvia
Steam switch
Setpoint: 95°, Steam 130°
Software: master (6177b2d6)

**How to reproduce**: Machine power is toggled while being at steam temperature.
Goes into `kEmergencyStop` and unable to leave, even if the actual temperature is below steam setpoint.
Only way to leave is cooling down to < 100° (brew setpoint + 5°)
PID_stopped

**Reason**:
- `EmergencyStopTemp` is initialized to 120°. `testEmergencyStop()` immediately triggers and machine goes to `kEmergencyStop`; `setEmergencyStopTemp()` will keep the StopTemp at 120 since `kSteam` is never reached.
- Leaving the mode compares against `brewSetpoint`, i.e. 95°. So even if the recognized (and displayed) setpoint in steam mode is higher than measured temperature, machine still displays "PID stopped" and refuses to heat.

**Solution**:
- Leaving kEmergency must compare against the current `setpoint` so the mode is left whenever the temperature goes near the current chosen setpoint.
- The initial EmergencyStopTemp should be the same 145 that are used in `setEmergencyStopTemp()`, to avoid setting `emergencyStop` immediately before first execution of `setEmergencyStopTemp()`.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -277,7 +277,7 @@ SysPara<float> sysParaScaleKnownWeight(&scaleKnownWeight, 0, 2000, STO_ITEM_SCAL
 
 // Other variables
 boolean emergencyStop = false;                // Emergency stop if temperature is too high
-double EmergencyStopTemp = 120;               // Temp EmergencyStopTemp
+double EmergencyStopTemp = 145;               // Temp EmergencyStopTemp
 float inX = 0, inY = 0, inOld = 0, inSum = 0; // used for filterPressureValue()
 boolean brewDetected = 0;
 boolean setupDone = false;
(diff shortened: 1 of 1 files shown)


kjyv (approved):

kjyv:
Note that this collides with https://github.com/rancilio-pid/clevercoffee/pull/488 (which will likely achieve the same result)

theHolgi:
Alright, I would retract my PR in favor of #488. But that one seems to be disputed?

kjyv:
Not in principle, only about the Quickmill topic

theHolgi:
Superseded by #488

PR #509, opened 2024-06-10
Add CurrentSensor class and update related files

lianuschka:
- Implemented an abstract class CurrentSensor.h for a generall current sensors
- Implemented analog sensor logic in CurrentSensorAnalog.cpp and CurrentSensorAnalog.h
- Updated pinmapping.h to include new sensor pins
- Modified main.cpp to initialize and use the new current sensor
- Updated MQTT functionality in mqtt.h for current value transmission
- Modified userConfig_sample.h to include current sensor options
- Improved timer utility in Timer.h, corrected a typo

### DIFF
--- a/src/hardware/CurrentSensor.h
+++ b/src/hardware/CurrentSensor.h
@@ -1,0 +1,149 @@
+/**
+ * @file CurrentSensor.h
+ *
+ * @brief Interface that all current sensors have to implement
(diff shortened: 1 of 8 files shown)


lianuschka:
Sorry, Falsches Repo. Versuche es zu Löschen.

PR #515, opened 2024-09-25
cleanup storage, variables and more

LoQue90:
no need to stay compatible to old storage structure anymore: 
-> remove reserved or freetouse 
rename a lot of vars to be consistent and follow camelcase

deleted unused coldstart stuff

set PID Mode based on pidPonmOn for regular and brew pid

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -180,7 +180,7 @@ void setBackflush(int backflush);
 void setScaleTare(int tare);
 void setScaleCalibration(int tare);
 void setNormalPIDTunings();
-void setBDPIDTunings();
+void setBrewPidTunings();
 void loopcalibrate();
 void looppid();
 void loopLED();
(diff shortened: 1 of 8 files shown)


src/main.cpp
```diff
@@ -1246,54 +1207,46 @@ void setup() {
...
-                                     .hasHelpText = true,
-                                     .helpText = F("The temperature that the PID will attempt to reach and hold"),
-                                     .type = kDouble,
-                                     .section = sTempSection,
-                                     .position = 11,
-                                     .show = [] { return true; },
-                                     .minValue = BREW_SETPOINT_MIN,
-                                     .maxValue = BREW_SETPOINT_MAX,
-                                     .ptr = (void*)&brewSetpoint};
+    editableVars["TEMP"] = {
```
kjyv, line 1216: why are some blocks formatted like before (e.g. for PID_PONM_ON above) and some like this inline? Not sure if this still needs to be auto-formatted

src/storage.h
```diff
@@ -66,97 +61,68 @@ typedef enum {
 
 // storage data structure
 typedef struct __attribute__((packed)) {
-        // Any 'freeToUse' areas ensure the compatibility to origin EEPROM layout and can be used by new value.
```
kjyv, line 69: Changing all this means that saved configurations will not work when updating firmware and Erase Flash needs to be used, correct?

src/main.cpp
LoQue90: i have no clue why this happens, but this is already formated. maybe @simonspa has an idea for it

src/storage.h
LoQue90: yes right.
i did talked about a 4.0 release with fiendie. There is the idea of postpone the new storage/config system to a 4.1 release.
If so, i would move this PR also to a 4.1 release, so there is only one point where all users have to do a flash erase and not two times.

src/main.cpp
simonspa: I think if the entire thing fits into the total line length defined, it will not be broken up, and otherwise aligned.

kjyv: If so then the line length might be configured to be too long? 220 chars usually doesn't fit on screens ;)

LoQue90: There was someone who wanted it so long, but it doesn't fit on my screen. (4k, only vsc with one side bar).
So I'm fine to break the line earlier.
Reading and editing the editiblevars isn't fun like that.

kjyv: Well, just "someone" shouldn't define the line lengths for the whole project ;) A more modern number might be 100 or 120, more traditional ones are 88.

LoQue90: hmm 120 does destroy the layout quit a lot, but formats all editibleVars as we want.
200 is still fine for my screen, editibleVars are still fine, but produce still some mess on other parts

PR #531, opened 2025-03-31
Update vertical template to support displayrotate instead of fixed orientation

Jaegerstefan:
A fixed rotation U8G2_R1 was hardcoded, but some vertical displays are installed in the opposite way, there U8G2_R3 is required. Use the user config settings to make it universal

### DIFF
--- a/src/display/displayRotateUpright.h
+++ b/src/display/displayRotateUpright.h
@@ -17,8 +17,9 @@ void u8g2_prepare(void) {
     u8g2.setDrawColor(1);
     u8g2.setFontPosTop();
     u8g2.setFontDirection(0);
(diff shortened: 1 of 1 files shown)


src/display/displayRotateUpright.h
```diff
@@ -17,9 +17,10 @@ void u8g2_prepare(void) {
     u8g2.setDrawColor(1);
     u8g2.setFontPosTop();
     u8g2.setFontDirection(0);
-    u8g2.setDisplayRotation(U8G2_R1); // fix setting for veritcal
+    u8g2.setDisplayRotation(DISPLAYROTATE); // either put U8G2_R1 or U8G2_R3 in userconfig
 }
 
+
```
fiendie, line 23: clang-format doesn't like the extra blank line :)

PR #536, opened 2025-04-25
Add button to reset WiFi settings

fiendie:


data/js/app.js
```diff
@@ -93,6 +93,22 @@ const vueApp = Vue.createApp({
...
+
+        async confirmReset() {
+            const confirmed = window.confirm(
+                "Are you sure you want to reset the WiFi settings?\n\nThis will erase saved credentials and restart the device."
+            );
+
+            if (!confirmed) return;
+
+            try {
+                const response = await fetch("/reset", { method: "POST" });
```
kjyv, line 106: Suggestion: This would probably be better called /wifiReset or similar, it's not resetting all configuration or only restarting the device.

kjyv (approved):
Apart from the rename, looks good :)

data/html/about.html
```diff
@@ -1,15 +1,31 @@
...
+    <script type="module">
+        if (window.appCreated === true) {
+            vueApp.mount('#content');
+        }
+        else {
+            window.addEventListener('appCreated', () => {
+                vueApp.mount('#content');
+            });
+        }
+    </script>
```
kjyv, line 12: The repetition can be prevented like this:
```suggestion
    <script type="module">
        (window.appCreated ? Promise.resolve() : new Promise(resolve => window.addEventListener('appCreated', resolve)))
            .then(() => vueApp.mount('#content'));
    </script>
```

PR #541, opened 2025-05-28
Fix PonM mode for PID controller

fiendie:
The PID controller was still using the P and I values of the old cold start mode which has since been removed. This commit refactors the method that sets the PID tunings so that it takes PonM into account when enabled.

### DIFF
--- a/src/storage.h
+++ b/src/storage.h
@@ -10,7 +10,7 @@
 // storage items
 typedef enum {
     STO_ITEM_PID_ON,                    // PID on/off state
-    STO_ITEM_PID_START_PONM,            // Use PonM for cold start phase (otherwise use normal PID and same params)
+    STO_ITEM_PID_USE_PONM,              // Use PonM mode for PID controller (Proportional on Measurement)
     STO_ITEM_PID_KP_START,              // PID P part at cold start phase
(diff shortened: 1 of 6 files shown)


src/storage.h
```diff
@@ -151,11 +151,11 @@ static const sto_data_t itemDefaults PROGMEM = {
...
-    STARTKP,                                                                                                                        // STO_ITEM_PID_KP_START
+    0,                                                                                                                              // free to use
     {0xFF, 0xFF},                                                                                                                   // free to use
     0,                                                                                                                              // STO_ITEM_SOFT_AP_ENABLED_CHECK
     {0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF},                                                                         // free to use
-    STARTTN,                                                                                                                        // STO_ITEM_PID_TN_START
+    0,                                                                                                                              // STO_ITEM_PID_TN_START
```
LoQue90, line 158: // free to use

LoQue90 (changes requested):
// editable vars are specified in main.cpp
#define EDITABLE_VARS_LEN 35

LoQue90 (approved):

PR #542, opened 2025-05-29
Timing optimisations

scottcondie24:
Changes have been made to minimise overlap of long running tasks and reduce excessive MQTT data. There is more that could be done but this only required minor changes instead of major rewrites. The boolean flags that block processes do increase the latency of these tasks by an extra loop, which is typically 0.5ms. Each long running task is given a unique flag, so blocking of other tasks can be selective on which tasks it will run with within the same loop.

Before
regular spikes over 100ms and even up to 2s when HASSIO discovery runs, as all the other processes were running in the same loop after this. These caused stuttery behaviour in the main screen and during brews and also increased the frequency of missed switch inputs.

After
These changes have resulted in the maximum loop time being typically 35ms (display sendBuffer) with very infrequent HASSIO discovery sends outside of critical times.

MQTT
Each time writeSysParamsToMQTT was called it would send around 70 messages for around 35 parameters, taking an average of 100ms each send. By storing the sent values in an array it reduces to only the changed variables, so send time is more like 15-50ms. Reducing the impact of these calls allows high speed refreshes during brews without significant increase to loop timing. The MQTT send interval has been split into Normal, Brew and Standby.

Display
sendBuffer() takes around 35ms to complete, and drawing the display into the buffer can take a few ms depending on its complexity. Changing to a boolean to flag the buffer is ready, then calling sendBuffer() on the next loop reduces the loop time to around 5ms then 35ms next loop, instead of one loop of 40ms. Standby mode with display off was still sending data to the display that was set to ignore it.

HASSIO discovery
Every 5 minutes HASSIO discovery was sent, taking between 200ms and 2s. If this happens during a brew the screen pauses for this time while the discovery message is sent. The retain flag is set so MQTT stores the discovery messages, so constant sending is not required. Limiting the send to only while not in a brew and only after MQTT disconnection reduces these blocking calls.

Website
The website refresh is only a few ms, blocking its update to only when other tasks have not run in the same loop minimises maximum loop time.

Standby
It would go into standby even if not enabled, so a check of standbyModeOn was added and if false then keep resetting the timers similar to resetStandbyTimer(void) but without the log.

Other changes
WeightSetpoint was not included in HASSIO discovery, and Steam Kp had an incorrect name.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -129,6 +129,15 @@ const unsigned long intervalPressure = 100;
 unsigned long previousMillisPressure; // initialisation at the end of init()
 #endif
 
+// timing flags
+bool displayBufferReady = false;
+bool displayUpdateRunning = false;
+bool websiteUpdateRunning = false;
+bool mqttUpdateRunning = false;
+bool hassioUpdateRunning = false;
+
+#include "utils/timingDebug.h"
+
 Switch* waterTankSensor;
 
 GPIOPin* statusLedPin;
@@ -627,6 +636,10 @@ void handleMachineState() {
             if (tempSensor->hasError()) {
                 machineState = kSensorError;
             }
+
+            if (machineState != kBrew) {
+                MQTTReCnctCount = 0; // allow MQTT to try to reconnect if exiting brew mode
+            }
             break;
 
         case kManualFlush:
@@ -793,6 +806,10 @@ void handleMachineState() {
                 u8g2.setPowerSave(0);
 #endif
                 machineState = kSensorError;
+            }
+
+            if (machineState != kStandby) {
+                MQTTReCnctCount = 0; // allow MQTT to try to reconnect if exiting standby
             }
             break;
 
@@ -1556,6 +1573,8 @@ void loop() {
 
     // Update LED output based on machine state
     loopLED();
+
+    debugTimingLoop();
 }
 
 void looppid() {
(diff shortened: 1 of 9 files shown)


src/main.cpp
```diff
@@ -151,6 +151,13 @@ const unsigned long intervalPressure = 100;
 unsigned long previousMillisPressure; // initialisation at the end of init()
 #endif
 
+// timing flags
+bool buffer_ready = false;
+bool display_update = false;
+bool website_update = false;
+bool mqtt_update = false;
+bool HASSIO_update = false;
+
```
LoQue90, line 160: this should follow camelCase ->
displayBufferReady
displayUpdateRunning
websiteUpdateRunning
mqttUpdateRunning
hassioUpdateRunning

scottcondie24:
To visually see the stutters I was changing the Hassio discovery time from 300000 (5mins) to 10000 (10s) then starting a full screen brew. The display shows time to 0.1s and it pauses often with the original code.

scottcondie24:
Debug functions have been added to print long loop times in the console with which activity caused it. This will aid in review but can be removed before approval if required

scottcondie24:
I added a fix for MQTT error counts not resetting, now it resets the counter to zero after 10mins since last connection, after waking from standby and after a brew is finished

src/main.cpp
```diff
@@ -1556,21 +1583,117 @@ void loop() {
 
     // Update LED output based on machine state
     loopLED();
+
+    debugTimingLoop();
+}
+
+// print what has caused the long loop time
```
fiendie, line 1590: I would probably want to move the enum and those methods to a separate header file. main.cpp is too long already :)

src/main.cpp
```diff
@@ -129,6 +129,22 @@ const unsigned long intervalPressure = 100;
...
 
+// timing flags
+bool displayBufferReady = false;
+bool displayUpdateRunning = false;
+bool websiteUpdateRunning = false;
+bool mqttUpdateRunning = false;
+bool hassioUpdateRunning = false;
+
+// Debugging - activity flags
+enum ActivityType : uint16_t {
```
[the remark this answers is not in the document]
scottcondie24: thanks, done

src/main.cpp
scottcondie24: moved all the timing debug code to a new file utils/timingDebug.h

scottcondie24:
Closing as I opened a new PR #561  based on the latest develop branch

PR #546, opened 2025-06-02
Added waterHandler code, removed brew(), added valve safety off

scottcondie24:
This code adds a waterHandler.h file to control the pump with the water switch.

It prioritises the switches as Brew, Steam then Water, so if the water switch is on then steam is activated it will run in kSteam. If brew is activated with either of the others active it will run in kBrew until completion.

Water switch is disabled during brew, enabled during Steam, disabled when water tank is empty.

There is debugging code embedded to flag what entry triggers the pump on or off, they can be removed.


scottcondie24:
Just tidied it up, functionally the same

scottcondie24:
Added extra check functions in brewHandler to remove brew() outside machineState. Added valve safety off in case it changes to an error state

src/main.cpp
```diff
@@ -62,6 +62,7 @@ enum MachineState {
     kBrew = 30,
     kManualFlush = 35,
     kSteam = 40,
+    kWater = 45,
```
LoQue90: kHotWater = 40
kSteam = 50
kBackflush = 60

src/waterHandler.h
LoQue90, line 1: i would like to do this like i did with the brew.
Having a state machine for the switch and a bool hotWater() that checks the state fo the switch state machine.
in the main state machine we call hotWater() in kPidnormal and kStandby.
checkWaterStates like checkBrewStates could then be used
and also a checkHotWaterActive()

src/main.cpp
```diff
@@ -707,6 +769,8 @@ char const* machinestateEnumToString(const MachineState machineState) {
             return "Manual Flush";
         case kSteam:
             return "Steam";
+        case kWater:
```
LoQue90: case kHotWater:
   return "Hot Water";
   
 
 and the order like suggested in the machine state enum

src/main.cpp
```diff
@@ -1195,6 +1267,7 @@ void loopPid() {
 
     checkSteamSwitch();
     checkPowerSwitch();
+    checkWaterSwitch();
```
LoQue90: checkHotWaterSwitch();

i would move that into bool hotWater()

src/main.cpp
```diff
@@ -1206,6 +1279,8 @@ void loopPid() {
 
     updateStandbyTimer();
     handleMachineState();
+    waterHandler();
```
LoQue90: hotWaterHandler();

LoQue90 (changes requested):
i would like to have this like the brewHandler :)

src/waterHandler.h
scottcondie24: Good idea, I initially wrote this like the steamHandler before your brew optimisations. I'll go through and rewrite it to be state based

LoQue90: I would like go go over steam as well and do the same. :)

src/main.cpp
fiendie: I think it's fine this way. The other "check" methods also have an early return. Keeps main.cpp a bit cleaner.

scottcondie24: I'm rewriting it all to hopefully be a bit cleaner

fiendie (approved):

PR #547, opened 2025-06-04
 Modernize Configuration Infrastructure

fiendie:
Storage Architecture Migration:
    - Replace EEPROM emulation layer with LittleFS JSON file storage
    - Migrate from SysPara functions to modern Config.h class with JSON persistence
    - Add hierarchical configuration structure with automatic validation

    Parameter System Refactor:
    - Replace SysPara with Parameter.h implementing type-safe Parameter classes
    - Support int, uint8, double, float, and string parameter types with validation
    - Implement automatic getter/setter binding with constraint enforcement
  
    Parameter Registry:
    - Replace editableVars with centralized ParameterRegistry system
    - Enable automatic config synchronization when parameters change
    - Provide unified interface for MQTT, web server, and internal parameter access

    Integration Updates:
    - Update MQTT parameter handling to use ParameterRegistry
    - Modify web server endpoints to use new parameter system

    Website Updates:
    - Add MQTT and System sections to parameter page 
    - Add Factory Reset on about page
    - Add config backup and restore feature to about page
    - Add restart button to about page

The main goal of this PR is to remove compile-time dependencies and make everything, including hardware-related settings, configurable at runtime. All config options are stored and read from config.json. Defaults and constraints are defined in the ConfigDef structure and defaults are stored in a map with the complete JSON path as the key, e.g. "hardware.switches.brew.enabled". ParameterRegistry serves as an additional layer on top of Config for web-configurable parameters. It adds information like section, help test, display name, etc. We now have a single point of truth for every setting.

Hardware-related settings are currently not web-configurable. We might add a separate hardware page to the web interface at some point but for the time being the onboarding process would involve the following steps: 
- Flash firmware and filesystem image (e.g. with ESP Flash Download Tool)
- Download default config.json from about page 
- Edit hardware-settings (leds, switches, sensors, etc.)
- Upload config.json
- Edit parameters (pid settings, setpoint, MQTT, etc.)


scottcondie24:
I tweaked my config.h file to

#include "userConfig.h"

            // MQTT
            _doc["mqtt"]["enabled"] = FEATURE_MQTT;
            _doc["mqtt"]["broker"] = MQTT_SERVER_IP;
            _doc["mqtt"]["port"] = MQTT_SERVER_PORT;
            _doc["mqtt"]["username"] = MQTT_USERNAME;
            _doc["mqtt"]["password"] = MQTT_PASSWORD;
            _doc["mqtt"]["topic"] = MQTT_TOPIC_PREFIX;
            _doc["mqtt"]["hassio"]["enabled"] = MQTT_HASSIO_SUPPORT;
            _doc["mqtt"]["hassio"]["prefix"] = MQTT_HASSIO_DISCOVERY_PREFIX;

I reactivated the LEDs that were commented out with no issues. I am able to download the config.json by navigating to silvia.local/download/config

src/Config.h
```diff
@@ -0,0 +1,1488 @@
...
+#pragma once
+
+#include "Logger.h"
+#include "defaults.h"
+#include "hardware/Relay.h"
+#include "hardware/Switch.h"
+#include <ArduinoJson.h>
+#include <LittleFS.h>
+
+class Config {
```
simonspa, line 19: On discord you discussed namespaces - which I personally think would be great to introduce - but I don't know what the status is wrt. to that since I have just started reading the PR changes. My suggestion would be to do that in a separate PR since adding a namespace (and its indentation) will generate a full-file diff.

src/Config.h
```diff
@@ -0,0 +1,1488 @@
...
+    public:
+        /**
+         * @brief Initialize the configuration system
+         *
+         * @return true if successful, false otherwise
+         */
+        bool begin() {
```
simonspa, line 26: I think I would move the initialization to the constructor. In the `main.cpp` you would then just keep a pointer around (`std::shared_ptr<Config> config`) and create the object when due `config = std::make_shared<Config>()` which initializes itself and loads defaults.

The benefit is that you cannot interact with the object without being initialized - something that can happen right now.

src/Config.h
```diff
@@ -0,0 +1,1488 @@
...
+                return save();
+            }
+
+            return true;
+        }
+
+        /**
+         * @brief Create a new configuration with default values
+         */
+        void createDefaults() {
```
simonspa: this is only called by `begin()` and should not be available to any other class dealing with a `Config` object, make it `private` (and possibly rename to `create_defaults()` to indicate the "private-ness".

src/Config.h
```diff
@@ -0,0 +1,1488 @@
...
+            if (!validateAndApplyConfig(doc)) {
+                return false;
+            }
+
+            return true;
+        }
+
+        // WiFi credentials flag
+        bool getWifiCredentialsSaved() {
+            return _doc["wifi"]["credentials_saved"] | false;
```
simonspa: I presume they are saved elsewhere (by the wifi manager?)?

src/Config.h
```diff
@@ -0,0 +1,1488 @@
...
+        void setPidEnabled(const bool value) {
+            _doc["pid"]["enabled"] = value;
+        }
+
+        bool getUsePonM() {
+            return _doc["pid"]["use_ponm"] | false;
+        }
+
+        void setUsePonM(const bool value) {
+            _doc["pid"]["use_ponm"] = value;
```
simonspa: An even cleaner solution than this would be to have a `PidConfig` class which has `getUsePonM()`, and `Config::getPidConfig()`

src/Config.h
```diff
@@ -0,0 +1,1488 @@
...
+
+        void setMqttHassioEnabled(const bool value) {
+            _doc["mqtt"]["hassio"]["enabled"] = value;
+        }
+
+        String getMqttHassioPrefix() {
+            return _doc["mqtt"]["hassio"]["prefix"];
+        }
+
+        void setMqttHassioPrefix(const String& value) {
```
simonspa: I am wondering if a more generic

```cpp
class Config {
  template <typename T>
  T get(const String& topic, const String& key) const;

  template <typename T>
  void set(const String& topic, const String& key, std::variant<String, bool, ...> , T& value);
  }
```

would not make this a cleaner interface

simonspa: Usage:

```cpp

auto myvalue = config.get<String>("mqtt","hassio_prefix");
myvalue += "_something";
config.set("mqtt", "hassio_prefix", myvalue);
```

src/Config.h
```diff
@@ -0,0 +1,1488 @@
...
+        void setScale2Calibration(const float value) {
+            _doc["hardware"]["sensors"]["scale"]["calibration2"] = constrain(value, -100000.0f, 100000.0f);
+        }
+
+    private:
+        inline static auto CONFIG_FILE = "/config.json";
+
+        StaticJsonDocument<4096> _doc;
+
+        bool validateAndApplyConfig(const JsonDocument& doc) {
```
simonspa, line 496: This file is *really* long and for the sake of legibility I would separate header and implementation. That makes it a lot easier to understand the structure of the class.

src/Config.h
```diff
@@ -0,0 +1,1488 @@
...
+            // PID parameters
+            if (doc["pid"].containsKey("enabled")) {
+                if (!doc["pid"]["enabled"].is<bool>()) {
+                    return false;
+                }
+
+                setPidEnabled(doc["pid"]["enabled"].as<bool>());
+            }
+
+            if (doc["pid"].containsKey("use_ponm")) {
```
simonspa: If I understand the code correctly, the JSON file is only ever accessed/created/altered through this class. IMHO it would then suffice to validate the parameters when setting and reading them instead of cycling through the entire file and re-implementing checks.

src/Config.h
```diff
@@ -0,0 +1,1488 @@
...
+                    return false;
+                }
+
+                setUsePonM(doc["pid"]["use_ponm"].as<bool>());
+            }
+
+            if (doc["pid"].containsKey("ema_factor")) {
+                const double value = doc["pid"]["ema_factor"].as<double>();
+
+                if (!validateParameterRange("pid.ema_factor", value, PID_EMA_FACTOR_MIN, PID_EMA_FACTOR_MAX)) {
```
simonspa: Example here, you use `constrain(/* ... */)` above and here again validate the constrained range - I think this is a bit duplicated

src/Config.h
```diff
@@ -0,0 +1,1488 @@
...
+                setPostBrewTimerDuration(value);
+            }
+
+            if (doc["display"].containsKey("heating_logo")) {
+                if (!doc["display"]["heating_logo"].is<bool>()) {
+                    return false;
+                }
+
+                setFeatureHeatingLogo(doc["display"]["heating_logo"].as<bool>());
+            }
```
simonspa: I am trying to understand what this function for example does. There are two options:

* `doc["display"]["heating_logo"]` **is** a boolean: The first condition returns false, function continues. We read the value of `doc["display"]["heating_logo"]` and assign it to `doc["display"]["heating_logo"]`. This is a no-op...?
* `doc["display"]["heating_logo"]` **is not** a boolean: first condition kicks in, function aborts, value stays a non-boolean JSON value.

What am I missing?

simonspa: oh, `doc` and `_doc`, now I see. Okay, still this might be a bit of a cumbersome validation to do for every parameter - and if the input is not correct, the value will not be set.

src/Parameter.h
```diff
@@ -0,0 +1,379 @@
...
+
+        [[nodiscard]] EditableKind getType() const {
+            return _type;
+        }
+
+        [[nodiscard]] int getSection() const {
+            return _section;
+        }
+
+        [[nodiscard]] int getPosition() const {
```
simonspa: I like the `[[nodiscard]]` sugar :)

src/Parameter.h
```diff
@@ -0,0 +1,379 @@
...
+    kDouble,
+    kDoubletime,
+    kCString,
+    kFloat,
+    kEnum
+};
+
+class Parameter {
+    public:
+        Parameter(
```
simonspa, line 18: I have not followed closely in the past months, but what has become of the idea to us a `class Parameter : public std::variant<std::int32_t, double, String, ...>`? I think it would implement some of the features already that this class provides

src/ParameterRegistry.cpp
```diff
@@ -0,0 +1,600 @@
+
+#include "ParameterRegistry.h"
+#include "Logger.h"
+#include "userConfig.h"
+#include <algorithm>
+
+ParameterRegistry ParameterRegistry::_singleton;
+
+// Global variables, needed for backwards compatibility
```
simonspa, line 8: Backwards to what? Older versions? How would they interfere here?
Backwards to still-existing legacy code in `main`?

src/ParameterRegistry.cpp
```diff
@@ -0,0 +1,600 @@
...
+    // Add all parameters
+    auto addParam = [this](const std::shared_ptr<Parameter>& param) {
+        _parameters.push_back(param);
+        _parameterMap[param->getId()] = param;
+    };
+
+    // PID Section
+    addParam(std::make_shared<Parameter>(
+        "PID_ON", "Enable PID Controller", kUInt8, sPIDSection, 1,
+        [&config]() -> bool {
```
simonspa: see comment in `Config` class. If we don;t re-implement setter and getter for individual keys there, using the parameter name (`PID_ON`) as the key it would be much simpler to access them without providing lambdas for accessing every single one of them

src/ParameterRegistry.cpp
```diff
@@ -0,0 +1,600 @@
...
+            usePonM = config.getUsePonM();
+            return usePonM;
+        },
+        [&config](const bool val) {
+            config.setUsePonM(val);
+            usePonM = val;
+        },
+        true, "Use PonM mode (<a href='http://brettbeauregard.com/blog/2017/06/introducing-proportional-on-measurement/' target='_blank'>details</a>)", [] { return true; }, &usePonM));
+
+    addParam(std::make_shared<Parameter>(
```
simonspa: I would suggest changing the `addParam` to accept variadic arguments and do the `make_shared` inside the function:

```cpp
template <typename... Args>
void addParam(Args&&... args) {
  _parameters.emplace_back(std::forward<Args>(args)...));
}
```

src/ParameterRegistry.cpp
```diff
@@ -0,0 +1,600 @@
...
+
+    _parameters.clear();
+    _parameterMap.clear();
+    _pendingChanges = false;
+    _lastChangeTime = 0;
+
+    // Add all parameters
+    auto addParam = [this](const std::shared_ptr<Parameter>& param) {
+        _parameters.push_back(param);
+        _parameterMap[param->getId()] = param;
```
simonspa: Do we need to store the parameter pointer twice?

simonspa: The `_parameters` vector is only returned in `getParameters()` but we could just build a vector from the map and return it.

src/ParameterRegistry.h
```diff
@@ -0,0 +1,216 @@
+
+#pragma once
+
+#include "Config.h"
+#include "Parameter.h"
+#include <map>
+#include <memory>
+#include <vector>
+
+enum ParameterSection {
```
simonspa, line 10: a parameter section seems to have an ID and a name - for me that woukld be enough to make it a `class` or at least ` struct`

simonspa (commented):
Alright, this is an enormous PR so I'm reading it in chunks. Here are my comments concerning the `Config`, `Parameter` and `ParameterRegistry` classes.

in my opinion, we should think this through a bit more on a architectural level. Right now, for my taste there are too many places in which we have specific setter and getter functions that need to know about value types and parameter names. Ideally, there is only one class that I ask about a variable and all the access code in registries and so on is independent of the type. See some ideas inline.

src/Config.h
fiendie: This is a bit of a relic. I kept it in place because some of the business logic in setup() was relying on it. Can probably be removed!

src/ParameterRegistry.cpp
fiendie: The latter, yes. We should probably move away from global variables in the long run.

src/display/displayTemplateManager.h
```diff
@@ -0,0 +1,62 @@
+#pragma once
+
+#include "displayCommon.h"
+
+namespace Templates {
+namespace Standard {
```
simonspa: I would like to put the namespaces into the respective headers instead of wrapping the `#include`.

simonspa:
```cpp
namespace Templates::Standard {
    /* ... */
}
```

should work

simonspa: Namespace indentation you should get with 

```diff
diff --git a/.clang-format b/.clang-format
index 8dd84fa..a2b2426 100644
--- a/.clang-format
+++ b/.clang-format
@@ -32,6 +32,7 @@ IndentCaseLabels: true
 IndentCaseBlocks: true
 IndentRequires:  false
 IndentWidth:     4
+NamespaceIndentation: All
 PackConstructorInitializers: NextLineOnly
 PointerAlignment: Left
 QualifierAlignment: Leave
```

To apply this to all files in the repo, run

```
$ pre-commit run --all-files
```

otherwise the new format will only be applied to the changed files.

simonspa: The reason it didn't work here previously likely is that `#include` is no code but a pre-processor macro, and they are never indented.

src/display/displayTemplateManager.h
```diff
@@ -0,0 +1,62 @@
...
+}
+
+namespace Scale {
+#include "displayTemplateScale.h"
+}
+} // namespace Templates
+
+class DisplayTemplateManager {
+    public:
+        enum Template {
```
simonspa, line 25: Hm, What I would do is not use namespaces but classes. You have a base `class Template` and then a `class Standard : public Template`. Then the manager instance does not need an additional `enum` but can directly say `currentPrintScreen = std::make_shared<T>()`

src/embeddedWebserver.h
```diff
@@ -86,72 +52,91 @@ String getTempString() {
...
-            case kInteger:
-                return String(*(int*)e.ptr);
-            case kUInt8:
-                return String(*(uint8_t*)e.ptr);
-            case kCString:
-                return String((char*)e.ptr);
-            default:
-                return F("Unknown type");
-                break;
+        const std::shared_ptr<Parameter> e = ParameterRegistry::getInstance().getParameterById(varName.c_str());
```
simonspa:
```suggestion
        const auto e = ParameterRegistry::getInstance().getParameterById(varName.c_str());
```

You don't have to type your fingers to the bone. :wink:

src/hardware/StandardLED.cpp
```diff
@@ -7,18 +7,16 @@
...
-            gpio.write(state ? LOW : HIGH); // Inverted logic
-        }
-        else {
-            gpio.write(state ? HIGH : LOW); // Normal logic
-        }
+    if (inverted) {
+        gpio.write(state ? LOW : HIGH); // Inverted logic
+    }
+    else {
+        gpio.write(state ? HIGH : LOW); // Normal logic
```
simonspa: This entire block should be equivalent to a XOR

```suggestion
        gpio.write(state != inverted ? HIGH : LOW);
```

src/mqtt.h
```diff
@@ -7,53 +7,75 @@
...
 
+inline void setupMqtt() {
+    ParameterRegistry& registry = ParameterRegistry::getInstance();
+
+    if (!registry.isReady()) {
+        LOGF(ERROR, "ParameterRegistry not ready, cannot initialize MQTT");
+        return;
+    }
+
+    mqtt_enabled = registry.getParameterById("MQTT_ENABLED")->getBoolValue();
```
simonspa: should have noted above, but you could template the function directly to make this

```cpp
    mqtt_enabled = registry.get<bool>("MQTT_ENABLED");
```

simonspa (commented):
Alright, that's the rest - overall really nice cleanup! :rocket:

LoQue90 (approved):

PR #548, opened 2025-06-06
Gzip compression

r4nd0wn:
- Moved static files from the data directory to the frontend directory to ensure the uncrompessed files are elsewhere than /data
- Added automatic compression steps to the PlatformIO build process, ensuring required frontend assets are compressed and copied to the data directory before building the filesystem.


r4nd0wn:
would need someone on a windows pc to test the filesystem build as i only have access to linux and mac....

scottcondie24:
> would need someone on a windows pc to test the filesystem build as i only have access to linux and mac....

I needed to add 
rel_file = rel_file.replace(os.sep, "/") 
after the line 
rel_file = os.path.join(rel_dir, file) if rel_dir != "." else file 
otherwise it would never match. After adding that line it successfully adds .gz into the existing folders

PR #549, opened 2025-06-06
Changed upright display to work with the new brew method

scottcondie24:
These are the changes I made to make the upright display usable. It follows the original display style and automatically adjusts based on pressure sensor and weight availability.

There are some improvements to make, like in the full screen timers.

There is a scrolling substring function in Common and RotateUpright that I'm using in my own build but not specifically used here.


scottcondie24:
closed as the code is stale

PR #559, opened 2025-06-25
Wifi changes, minor tweaks to upright display, removed boolean for invert display

scottcondie24:
displayTemplateUpright
fixed water icon not showing when full screen brew or flush not selected
added MQTT
fixed when BREW or FLUSH was displayed

displayCommon
Minor shift of water icon
removed boolean for display inverted

ParameterRegistry.h
removed boolean for display inverted

### DIFF
--- a/src/display/displayTemplateUpright.h
+++ b/src/display/displayTemplateUpright.h
@@ -25,7 +25,6 @@ inline void printScreen() {
         return;
     }
 
-    // If no specific machine state was printed, print default:
     u8g2->clearBuffer();
     if (machineState == kWaterTankEmpty) {
         u8g2->drawXBMP(8, 50, Water_Tank_Empty_Logo_width, Water_Tank_Empty_Logo_height, Water_Tank_Empty_Logo);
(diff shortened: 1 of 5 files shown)


src/display/displayTemplateUpright.h
```diff
@@ -101,18 +95,14 @@ inline void printScreen() {
...
-                    }
-                    else if (shouldDisplayBrewTimer()) {
-                        u8g2->print("BREW");
-                    }
-                    else {
-                        u8g2->print("OK");
-                    }
-                }
+            if (featureBrewControl && machineState == kManualFlush) {
+                u8g2->print("FLUSH"); // no flush function if optocoupler? manual flush is still triggered if momentary
```
LoQue90: Manual flush only works with momentary switch.
Optocoupler should be used with toggle.

scottcondie24: Thanks, removed old comment. Function only works if brew switch and brew control is enabled and in flush mode

scottcondie24:
Added changes to wifi connection and reconnection so it isn't blocked while attempting to reconnect

LoQue90 (dismissed):

xX7:
> Added changes to wifi connection and reconnection so it isn't blocked while attempting to reconnect

Was just about to think about a solution for that as well. Had a MAJOR issue with the blocking wifi reconnect, which led to an overheat prevented by the thermo-fuse (quickmill), as the PID is not updated and in my case left the SSR on.

scottcondie24:
> > Added changes to wifi connection and reconnection so it isn't blocked while attempting to reconnect
> 
> Was just about to think about a solution for that as well. Had a MAJOR issue with the blocking wifi reconnect, which led to an overheat prevented by the thermo-fuse (quickmill), as the PID is not updated and in my case left the SSR on.

I had a few events like that until I modified my antenna, but I saw a few people were having the same issue so figured I should fix it at the source

src/main.cpp
```diff
@@ -314,45 +315,51 @@ void initOfflineMode() {
  * @brief Check if Wifi is connected, if not reconnect abort function if offline, or brew is running
  */
 void checkWifi() {
+    static int count = 1;
+    static bool statusTemp = false;
```
LoQue90: Maybe does could be more specific.

LoQue90 (approved):

src/main.cpp
scottcondie24: Makes sense, I just reused the existing naming but can easily modify

LoQue90: Always using the chance for improvements :D

scottcondie24: done, also removed yield() as the function is non blocking

PR #561, opened 2025-06-28
Timing optimisations to reduce stutter

scottcondie24:
Added process flags to reduce stuttering
Added MQTT filter to only send what has changed
Display updates in main screens are split into a draw loop then sendBuffer on the next loop
Block sending HASSIO discovery updates during critical tasks

Added tickbox to system menu to turn on and off the change (except splitting display), and added a tickbox to remove display updates from the logs to reduce frequency of sending log data.

The best way to test is to use the process limit tick box and compare how smooth the display updates are in a brew etc.
MQTT is a major source of stutters without the process limit, so you see pauses every 5 seconds.

This is an improved version from #542 

I expect some of the functions I added for testing the improvements could be removed once reviewed.

### DIFF
--- a/src/ParameterRegistry.cpp
+++ b/src/ParameterRegistry.cpp
@@ -48,6 +48,9 @@ extern bool scaleTareOn;
 extern bool scaleCalibrationOn;
 extern int logLevel;
 extern const char sysVersion[64];
+extern bool timingMaster;         // if set to false it lets processes run together, if true it limits how many processes run per loop
+extern bool includeDisplayInLogs; // if set to true it shows every display update, if false it only shows if > 45ms or if running in the same loop as another process
+extern bool timingDebugActive;
 
 void ParameterRegistry::initialize(Config& config) {
     if (_ready) {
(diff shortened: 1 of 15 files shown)


scottcondie24:
split mqtt into chunks of 10ms if process limit is enabled. This is seen by the consecutive M results in the console logs

LoQue90 (approved):
after removing the debug part, i do like it!

src/ParameterRegistry.cpp
```diff
@@ -691,6 +694,41 @@ void ParameterRegistry::initialize(Config& config) {
...
+            sSystemSection,
+            1111,
+            &timingDebugActive,
+            "Enable or disable the process loop time debugging in console.<br>"
+            "r=draw display buffer<br>"
+            "D=display refresh<br>"
+            "W=website<br>"
+            "M=MQTT<br>"
+            "H=hassio<br>"
+            "T=temperature"
```
fiendie: You can move the check for the log level into showCondition: 

`, [&config] { return config.get<int>("system.log_level") == static_cast<int>(Logger::Level::DEBUG); }`

fiendie (approved):

PR #563, opened 2025-06-28
Use VERSION.txt instead of hard-coded values in main.cpp

fiendie:


stan23:
The branch name could be added as well:

`from dulwich.porcelain import active_branch`

`branch_name = active_branch(r).decode("utf-8")`

fiendie:
> The branch name could be added as well:
> 
> `from dulwich.porcelain import active_branch`
> 
> `branch_name = active_branch(r).decode("utf-8")`

Done :)

LoQue90 (approved):

PR #566, opened 2025-06-30
Change loglevel on the fly

stan23:
no need to reboot after changing loglevel in webinterface


stan23:
sorry, with all the rebasing/squashing the whitespace sneaked in again and violated clang again

LoQue90 (approved):

PR #567, opened 2025-06-30
Create abstraction layer and add support for BLE scales

fiendie:
* Added abstract Scale class
* Moved hard-coded HX711 implementation from scaleHandler to separate class
* Added initial support for Bluetooth Low Energy scales based on a fork of the AcaiaArduinoBLE library


scottcondie24:
I see you made a brew states file full of enums, should there just be a combined enums file for everything?

fiendie:
> I see you made a brew states file full of enums, should there just be a combined enums file for everything?

Yeah, maybe :D I had to move the states to a separate header file because brewHandler and scaleHandler would have had a circular dependency otherwise.

src/hardware/HX711Scale.cpp
```diff
@@ -0,0 +1,182 @@
...
+        if (!readSecondScale) {
+            if (loadCell1->update()) {
+                weight1 = loadCell1->getData();
+                readSecondScale = true;
+            }
+        }
+        else {
+            if (loadCell2->update()) {
+                weight2 = loadCell2->getData();
+                currentWeight = weight1 + weight2;
```
scottcondie24: I would move this to the end of the else statement or add it to loadcell1, then it gives the latest total weight as fast as possible instead of after the second cell is read

src/scaleHandler.h
```diff
@@ -30,72 +57,52 @@ inline void scaleCalibrate(HX711_ADC loadCell, const int pin, const bool isSecon
...
-    loadCell.setSamplesInUse(scaleSamples);
+    loadCell->setSamplesInUse(128);
+    loadCell->refreshDataSet();
+    float calibration = loadCell->getNewCalibration(scaleKnownWeight);
+    loadCell->setSamplesInUse(scaleSamples);
 
-    LOGF(INFO, "New calibration: %f", *calibration);
+    LOGF(INFO, "New calibration: %f", calibration);
 
     u8g2->sendBuffer();
```
scottcondie24: I missed this one, it is not needed

src/hardware/HX711Scale.cpp
```diff
@@ -0,0 +1,182 @@
...
+    if (cellNumber == 1) {
+        return loadCell1;
+    }
+    else if (cellNumber == 2 && isDualCell) {
+        return loadCell2;
+    }
+    return nullptr;
+}
+
+void HX711Scale::calibrateCell(HX711_ADC* cell, int cellNumber, float knownWeight) {
```
scottcondie24: this never gets used

src/hardware/HX711Scale.cpp
```diff
@@ -0,0 +1,182 @@
...
+    loadCell1->tare();
+    loadCell1->setCalFactor(calibrationFactor1);
+
+    if (isDualCell) {
+        loadCell2->tare();
+        loadCell2->setCalFactor(calibrationFactor2);
+    }
+}
+
+bool HX711Scale::startCalibration(float knownWeight) {
```
scottcondie24: this never gets used

src/hardware/HX711Scale.cpp
```diff
@@ -0,0 +1,182 @@
...
+void HX711Scale::calibrateCell(HX711_ADC* cell, int cellNumber, float knownWeight) {
+    float& calibrationFactor = (cellNumber == 1) ? calibrationFactor1 : calibrationFactor2;
+
+    cell->setCalFactor(1.0);
+    cell->update();
+    cell->tare();
+
+    delay(10000);
+
+    const int originalSamples = 16; // HX711_ADC default
```
scottcondie24: samples should be reset to same as before calibration
const int originalSamples = cell->getSamplesInUse();

src/scaleHandler.h
```diff
@@ -6,18 +6,45 @@
```
scottcondie24, line 6: this file has a lot of HX711 references, if it is being abstracted should this be changed?

scottcondie24 (changes requested):
The code works, but if bluetooth gets added as a scale alternative then the scaleHandler would need another rewrite.

src/scaleHandler.h
```diff
@@ -124,90 +241,120 @@ inline void checkWeight() {
...
-        byte loadCell2Ready = 0;
-
-        // run startup, stabilization and tare, both modules simultaniously
-        // this parallel start seems to be the most important part to get accurate readings with two HX711s connected
-        while (loadCellReady + loadCell2Ready < 2) {
-            if (!loadCellReady) {
-                loadCellReady = LoadCell.startMultiple(stabilizingtime, _tare);
-            }
+        // HX711 scale types
+        const float cal1 = scaleCalibration;
```
scottcondie24: these global variables are no longer used, so they default to 1.
Use hardware.sensors.scale.calibration and hardware.sensors.scale.calibration2 then the references to these variables can be deleted from main and parameterregistry.cpp

src/scaleHandler.h
```diff
@@ -1,23 +1,165 @@
...
+                    const bool brewByWeightEnabled = config.get<bool>("brew.by_weight");
+                    const bool brewByTimeEnabled = config.get<bool>("brew.by_time");
+
+                    if (brewByWeightEnabled && brewByTimeEnabled) {
+                        LOG(INFO, "Activating brew-by-time fallback due to scale connection loss");
+                        brewByWeightFallbackActive = true;
+                    }
+                    else if (brewByWeightEnabled) {
+                        LOG(WARNING, "Scale connection lost during brew-by-weight only mode - setting weight to target");
+                        currBrewWeight = targetBrewWeight; // This will trigger brew completion
```
scottcondie24: currBrewWeight gets written often so I have seen this not trigger. Directly modifying the brew state works currBrewState = kBrewFinished;  or alternatively changing the targetBrewWeight but that may impact user settings. If you still want to modify a weight you could minus the targetBrewWeight from the preBrewWeight. The log wording may need to change to reflect what happens. In brewHandler these stop states are checking if time or weight are >, not >=.

src/ParameterRegistry.cpp
```diff
@@ -378,44 +378,9 @@ void ParameterRegistry::initialize(Config& config) {
             },
```
scottcondie24, line 384: extern float scaleCalibration;
extern float scale2Calibration;
extern float scaleKnownWeight;

dont have references below, as they are all nullptr. If not used then they need to be removed from main and scaleHandler too

Here are some defaults to add to the entries below
#define SCALE_CALIBRATION_MIN         -999999.0
#define SCALE_CALIBRATION_MAX         999999.0
#define SCALE2_CALIBRATION_MIN        -999999.0
#define SCALE2_CALIBRATION_MAX        999999.0
#define SCALE_KNOWN_WEIGHT_MIN        1.0
#define SCALE_KNOWN_WEIGHT_MAX        2000.0

src/display/displayCommon.h
```diff
@@ -382,6 +382,14 @@ inline void displayProgressbar(const int value, const int x, const int y, const
     }
 }
 
+inline void displayBluetoothStatus(const int x, const int y) {
```
scottcondie24, line 386: this isn't called for the upright template (top right corner should be free), and interferes with the reconnect counter when wifi disconnects.
It shows when scale is changed to bluetooth in hardware webpage, though a restart fixes it

src/hardware/scales/BluetoothScale.cpp
```diff
@@ -0,0 +1,126 @@
...
+        return false;
+    }
+
+    if (connected) {
+        if (bleScale->heartbeatRequired()) {
+            bleScale->heartbeat();
+        }
+
+        if (bleScale->newWeightAvailable()) {
+            if (const float newWeight = bleScale->getWeight(); newWeight >= 0 && newWeight < 10000) {
```
scottcondie24: If tare is on -0.1 or lower it freezes updates at the last value >=0. When a brew starts it gets currReadingWeight which would be this value >=0,
For example if 11g was the last read value before -0.1 then a brew will start and zero at 11g, so the scales then need to measure > 11g before it starts counting up.
It might be better to let through some negative values then it behaves like the hx711

src/hardware/scales/HX711Scale.cpp
```diff
@@ -0,0 +1,158 @@
...
+            weight1 = loadCell1->getData();
+            currentWeight = weight1;
+            return true;
+        }
+    }
+    else {
+        if (!readSecondScale) {
+            if (loadCell1->update()) {
+                weight1 = loadCell1->getData();
+                readSecondScale = true;
```
scottcondie24, line 90: currentWeight = weight1 + weight2;

src/scaleHandler.h
```diff
@@ -124,90 +241,120 @@ inline void checkWeight() {
...
-
-        // run startup, stabilization and tare, both modules simultaniously
-        // this parallel start seems to be the most important part to get accurate readings with two HX711s connected
-        while (loadCellReady + loadCell2Ready < 2) {
-            if (!loadCellReady) {
-                loadCellReady = LoadCell.startMultiple(stabilizingtime, _tare);
-            }
+        // HX711 scale types
+        const float cal1 = scaleCalibration;
+        const float cal2 = scale2Calibration;
```
scottcondie24: these still default to 1, so I need to calibrate on every reboot

src/scaleHandler.h
```diff
@@ -1,23 +1,165 @@
...
+ */
+inline bool shouldUseBrewByWeight() {
+    const bool brewByWeightEnabled = config.get<bool>("brew.by_weight");
+    return brewByWeightEnabled && !brewByWeightFallbackActive && !scaleConnectionLost;
+}
+
+inline void scaleCalibrate(const int cellNumber, const int pin) {
+    if (isBluetoothScale) {
+        // Bluetooth scales handle calibration internally
+        displayWrappedMessage("Bluetooth scales\nhandle calibration\ninternally");
```
scottcondie24, line 141: langstring

unverified translations
Bluetooth-Waagen übernehmen die Kalibrierung intern
Las balanzas Bluetooth manejan la calibración internamente

scottcondie24 (changes requested):
I have added all the issues I found as single comments

src/hardware/HX711Scale.cpp
fiendie: I'm not quite sure what you mean

src/scaleHandler.h
fiendie: This is never actually displayed because I hide the calibrate button when scaleType is 2 lol

src/scaleHandler.h
fiendie: The log statement?

scottcondie24: the u8g2->sendBuffer(); can be deleted, its all done in displayWrappedMessage(msg);

src/display/displayCommon.h
scottcondie24: This still interferes with the RC: 1 messages when wifi disconnects. The RC could be moved to the right as MQTT will not be connected if wifi drops out

src/scaleHandler.h
scottcondie24: oh I recall seeing that now, that makes sense

scottcondie24:
there is still the overlap of the bluetooth symbol over the reconnect counter RC: 1

PR #441, opened 2024-01-20
Start dimmer experiments

FabianSperrle:
This is an initial attempt at implementing pump control with the Robotdyn dimmer.
Currently, pump pulses are very irregular in their timing. My expectation was that I would get a constant, low-frequency pump motion.

Leaving this here as a (hopefully useful) basis for discussion, a starting point for other experiments, and to get feedback from everybody.

PR #576, opened 2025-07-14
Dimmer code and brew profiles

scottcondie24:
This code is a work in progress

It contains dimmer code that allows setting two different control techniques, pulse skip modulation (PSM) like Decent DE1 and phase modulation (Phase) like the leva! system. PSM works best when using flow rates, Phase flow calculations still need adjustment.

There is a pump PID controller that works for either a pressure or flow target. There is code to help limit pressure while in flow mode

There are brew profiles that control the pump during brew, I created them based on a lot of reading and watching videos of how different lever machines worked, comparing against my own lever machine, and seeing what types of profiles people enjoy on pressure profiling machines. I have tested most of them and enjoy the Londinium Vectis most for the beans I use. I have some pressure only profiles for use with the Phase dimmer, though there is no code to stop any selections.

Some of the profiles are intended to be stopped manually, where the last phase is long so you can stop at your desired volume. The Adaptive profile works well when you are unsure of grind size/dose, as it limits to near 9 bars if too fine, and has a maximum flow rate if too coarse. I need to tweak the transitions on the Pressurised Bloom and increase time on Blooming Espresso.

I added a graph page on website, it activates at the start of a brew and sends data in 100ms intervals.

In the debug console there is the option to print arrays of brew data, sending 9 arrays every 2 seconds, each containing 20 values at around 50ms intervals.

The Scale display template had an extra data section down the bottom if a dimmer and pressure sensor are enabled.
It shows current pressure, current power, the control mode, and on the next line the flow rate and the control mode data or profile name that changes to brew phase name during a brew.

I excluded the encoder code as I did not create a menu, all variables can be changed on the website.

There are changes to the webpage source data so you need to upload filesystem as well as the main code.

HARDWARE REQUIREMENTS
The dimmer code requires a zero crossing signal, so if only a relay is installed it will not try to control the output.
I use a 4A robotdyn dimmer
A pressure sensor is required for pressure and flow control, though without one you could still use the power mode
Some profiles include weights, if no scales are present it defaults to the target time instead for that phase

This includes concepts from the start dimmer experiments #441 published by @FabianSperrle


scottcondie24:
I have made a lot of changes to the phase controller, linearising the power input to flow output, adding a pressure scaler to compensate for needing high power for any flow at high pressure, and adding linear interpolation to the phase delay code to remain smooth. Phase control is still inherently less accurate at flow than the Pulse Skip but it is now a lot closer.
To test and initially tune the PID I use a blind basket with the portafilter not completely engaged, providing some pressure and a little flow. Use the graph page on a phone and vary the target on another device then check how it responds to the changed target.

scottcondie24:
I added profile and phase text descriptions on the graph page to make it more user friendly. It now gives an indication whether it will automatically stop or if it is intended for the user to manually stop at their desired volume

scottcondie24:
I fixed the graph page after the latest rebase issues. There is very low heap space available on the current hardware so it can be helpful to disable some functions like MQTT while initially loading the graph page. Once the browser has cached the required files performance improves

scottcondie24:
I updated to the latest develop and removed the bug of restoring the last boiler setpoint when changing modes

scottcondie24:
I enabled the automatic brew by time and brew by weight user settings into the profile stop checks. This may need some warnings if a low brew time is selected on certain profiles

scottcondie24:
Rebased to latest master and added export buttons of latest brew to graph page

scottcondie24:
There are a few big changes in this one

Temperature has been added to the graph so you can monitor it during brew. 100ms updates help when tuning the brew PID variables.

The graph no longer needs a dimmer to activate. If no dimmer is present it sets the pressure and flow targets to zero but allows graphing of temperature, pump power, pressure, flow rate through hardware sensor, and brew weight. Exporting to json or CSV is also available.

A hardware flow sensor has been added on pin 14, this interferes with JTAG so cannot be included in more complex debugging. Digmesa nano DM60 and it's equivalents are supported, it has significantly more pulses per mL than others and starts operating at lower flow rates. There is a tick box in the debug menu to enable printing pulse data to the console. To view flow sensor data on the display use the Scale template if a dimmer and pressure sensor is enabled, otherwise it can be seen during brew on the graph page. If a hardware flow sensor is not connected the profiles still work using software calculations, though depending on the control type it is less accurate. Flow PIDs have been tuned for a stock pump with software calculated flow rate, not the hardware flow sensor.

I changed the frequency selector to an auto frequency detection method.

Beta velofuso support has been added. 

Hardware requirements are a velofuso DC gear pump, replace the potentiometer with the output of an mcp4725 connected at 5V. The mcp4725 takes an i2c input, to connect needs a logic level shifter to convert the existing 3.3v i2c to 5v.

The software includes a selection of DAC Velofuso (Beta) in the dimmer type drop down. The PID values are available but have not been tuned. There is a preliminary software flow rate calculation but as I do not have one for testing it is recommended to use the hardware flow sensor.

I also modified the pins reserved for the encoder, pin 3 interferes with serial data so is least impacted by connecting to a switch.

PR #580, opened 2025-07-22
added a step to config portal start to send the correct password, move RC Upright

### LINKED ISSUE #323: After first time setting WiFi credentials and connecting, Webserver is not reachable
When setting up a new machine, the wifi credentials can be entered and the esp connects to it and displays an IP in the the log. However, connecting to it and showing the webinterface is only possible when restarting the ESP

scottcondie24:
added a short 1s timeout so the ConfigPortal receives and stores hostname and password. Timeout is reset to 60s and ConfigPortal restarted and then the display is updated to say Starting Portal AP

### DIFF
--- a/src/display/displayCommon.h
+++ b/src/display/displayCommon.h
@@ -97,7 +97,14 @@ inline void displayWiFiStatus(const int x, const int y) {
     }
     else {
         u8g2->drawXBMP(x, y, 8, 8, Antenna_NOK_Icon);
-        u8g2->setCursor(x + 36, y - 1);
+
+        if (config.get<int>("display.template") == 4) {
+            u8g2->setCursor(x + 12, y - 1);
+        }
+        else {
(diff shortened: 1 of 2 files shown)


scottcondie24:
I added a restart as I was never able to directly access silvia.local if I didn't restart after the wifi configuration

LoQue90:
> I added a restart as I was never able to directly access silvia.local if I didn't restart after the wifi configuration

Great, I think I that's something we had somewhere on the road map or in an issue.

scottcondie24:
Moving the reconnect counter away from the bluetooth symbol for most displays put it behind the bluetooth symbol on the upright template, this fixes it

fiendie:
> > I added a restart as I was never able to directly access silvia.local if I didn't restart after the wifi configuration
> 
> Great, I think I that's something we had somewhere on the road map or in an issue.

It's issue #323

fiendie (approved):

PR #582, opened 2025-07-28
refactor: adjust parameter and configuration to be stored in NVS pers…

BlackDark:
I noticed a problem during testing of new UI implementations that after each filesystem change the whole configuration gets wiped (makes sense because the configuration was stored as file json on the FS).

I reimplemented quite a bunch to store the configuration parts and as before sync back to global vars with the NVS system.
Now the configuration persists after each flash.

Not fully finished yet, have to extract some things.


fiendie:
This is more like a feature, not a bug. The config only gets wiped if the filesystem image is written and you can just make a backup of your config beforehand. 

I don't see this global variable approach as much of an improvement to be honest. We're trying to get rid of those in the near term. And the two-tier approach with Config and ParameterRegistry is by design. 

Sorry you've put this much effort into this.

BlackDark:
No worries, was with the help of copilot so wasn't so much time tinkering around :)
Got to learn a bit about old/unpretty languages like C++  :D

Feel free to use it as idea or possible motivation for your planned future refactorings 👍

PR #584, opened 2025-07-29
Added streaming function for parameters, optimised fonts, fixed scale crash

scottcondie24:
This feature regularly clears memory and reduces crashes, even when available memory is low.
Added defer to .js headers to reduce concurrent access requests to ESP32.
The biggest improvement is seen on mobile where the Settings and Hardware pages load more reliably.

This has only been tested on Android with Chrome and Samsung Internet browsers.

### DIFF
--- a/frontend/js/app.js
+++ b/frontend/js/app.js
@@ -7,7 +7,7 @@ const vueApp = Vue.createApp({
             parametersHelpTexts: [],
             isPostingForm: false,
             showPostSucceeded: false,
-            filter: 'behavior',
+            filter: '',
 
             // Config upload properties
             selectedFile: null,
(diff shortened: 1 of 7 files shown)


scottcondie24:
Using 
& "$env:USERPROFILE\.platformio\packages\toolchain-xtensa-esp32\bin\xtensa-esp32-elf-nm.exe" -S --size-sort -t d .pio/build/esp32_usb/firmware.elf | Select-String " [BD] "
In a terminal window in VSCode it showed fonts used a lot of flash space. Reducing some of them down saved over 36KB without any visual impact.
tf has all characters including accents
tr is reduced with only standard letters and numbers
tn is numbers and a few symbols

fiendie (approved):

Issue #539, opened 2025-05-16
possibility to deactivate flashing temperature - feature request


Scharnhorst74:
Dear CleverCoffee-Team,

would it be possible to decide in the userconfig by the enduser, wether the temperature in the display template "temperature only" is flashing or not?
Actually the temperature is flashing when the target temperature is reached.

Thank you in advance for your support!

Best regards,

Paddy.

LoQue90:
just a workaround until i have a good idea how to make this possible the change on runtime

https://github.com/LoQue90/clevercoffee/tree/temponly_blinking

LoQue90:
add in current 4.0 beta 3

PR #590, opened 2025-08-28
Bugfixes Beta3

scottcondie24:
I added a css and js loader which attempts to access local and remote files at the same time, then cancels the local if remote loads first. They mostly load one at a time reducing strain on the ESP, and  the developer tools console log shows which version loads to confirm if it was local or CDN.
I added a warning if page doesn't load completely, but this still may not trigger if certain components fail.
I increased pid maxs as I was at the limit of Tn and wanted to try further.
I added blink parameters so now it is more customisable, it can be disabled to suit the issue #539 
The pressure sensor now returns the last known value if it detects disconnection without waiting for the interface to timeout
The temperature sensor would block the display from updating if loops were too long, so I added force display update on long loops.
CONFIG_REFERENCE was updated with a few missing parameters.
Offline mode now starts a softAP and starts the server so configuration is still possible, I added a parameter registry entry now that Offline mode can be toggled on and off without reflashing the device.
I changed a startup delay that now happens if no scale or if bluetooth scale is connected, giving time to read the IP address on boot.

### DIFF
--- a/CONFIG_REFERENCE.md
+++ b/CONFIG_REFERENCE.md
@@ -119,6 +119,11 @@ This document describes all configuration parameters available in the `config.js
 - **Default**: `false`
 - **Description**: Show manual flush timer in fullscreen mode
 
+### `display.fullscreen_hot_water_timer`
+- **Type**: Boolean
+- **Default**: `false`
+- **Description**: Show hot water timer in fullscreen mode
+
 ### `display.heating_logo`
 - **Type**: Boolean
 - **Default**: `true`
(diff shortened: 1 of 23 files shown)


scottcondie24:
Changed how it displays the softAP information and fixed a bug that made it unusable in some situations due to HASSIO attempting to send.
Fixed decimal place displays for temperatures >= 100

scottcondie24:
I changed the mqtt file to reduce the potential for heap fragmentation, and as a bonus it reduced the flash usage a little.
MachineState now sends an enum string to mqtt.
Number2string has been removed.
There were also some langstring tweaks and I removed the IFLOG arguments as logger handles it.

fiendie (approved):

PR #592, opened 2025-09-01
Don't re-register OTA callbacks, fence off MDNS/OTA during WiFi-reconnects

fiendie:


src/main.cpp
```diff
@@ -1592,3 +1612,32 @@ void setBDPIDTunings() {
...
+        if (MDNS.begin(hostname.c_str())) {
+            mdnsStarted = true;
+        }
+        else {
+            LOG(ERROR, "Error setting up MDNS responder!");
+        }
+    }
+
+    if (!otaStarted) {
+        ArduinoOTA.setHostname(hostname.c_str());
```
scottcondie24: does password need to be set here? It gets called before the OTA initialisation around line 1032

src/embeddedWebserver.h
```diff
@@ -423,8 +423,16 @@ inline void serverSetup() {
...
-        const String json = getTempString();
-        request->send(200, "application/json", json);
+        AsyncResponseStream* response = request->beginResponseStream("application/json");
+        response->print('{');
+        response->print("\"currentTemp\":");
+        response->print(curTemp, 6);
+        response->print(",\"targetTemp\":");
+        response->print(tTemp, 6);
+        response->print(",\"heaterPower\":");
+        response->print(hPower, 6);
```
scottcondie24: these could be reduced to 1 or 2 decimal places
{"currentTemp":19.5,"targetTemp":25,"heaterPower":39.03606534}

src/main.cpp
```diff
@@ -1005,12 +1011,37 @@ void setup() {
         wiFiSetup();
         serverSetup();
 
+        WiFi.onEvent([](WiFiEvent_t e, WiFiEventInfo_t) {
```
scottcondie24: This is defined after wifi is already connected, so it doesn't trigger an event and startMdnsOta never gets called

scottcondie24 (changes requested):
I am looking at what needs to change but currently OTA is not working on my ESP, when I added a LOG( the startMdnsOta never gets called, and I can only access the website with my IP not mDNS hostname

scottcondie24:
I moved WiFi.onEvent( above wiFiSetup() and all OTA config into the startMdnsOta() function and it works, though not sure if moving all the OTA was needed

src/main.cpp
```diff
@@ -1008,9 +1029,6 @@ void setup() {
         // OTA Updates
         if (WiFi.status() == WL_CONNECTED) {
             otaPass = config.get<String>("system.ota_password");
```
scottcondie24: otaPass should be set at the top of configureArduinoOta otherwise OTA is done when otaPass has not been set

src/main.cpp
```diff
@@ -788,6 +794,21 @@ void wiFiSetup() {
     wm.setDisableConfigPortal(true); // disables config portal on wifi save
     bool wifiConnected = wm.autoConnect(hostname.c_str(), pass);
```
scottcondie24: this should go below wifi event

scottcondie24 (changes requested):
Timeseries worked well and after moving those two lines in my test it all works well. I am not able to test mDNS related crashes but I can see it still works for OTA and website access

scottcondie24:
I kept getting errors related to multiple calls to MDNS.end, ArduinoOTA closes a connection already. I'm currently testing with these changes but will test for a few more days.

void stopMdnsOta() {
    if (otaStarted) {
        ArduinoOTA.end();
        otaStarted = false;
        mdnsStarted = false;    //ArduinoOTA calls mdnsSocket.stop();
    }
    else if (mdnsStarted) {
        MDNS.end();
        mdnsStarted = false;
    }
}

scottcondie24:
I just noticed in my code I also had this commented out in the wifi reconnect section. I don't recall if I tested both versions

if (wifiConnectCounter == 1) {
                wifiReconnects++;
                LOGF(INFO, "Attempting WIFI (re-)connection: %i", wifiReconnects);
                //stopMdnsOta();
                wm.disconnect();
                WiFi.begin();
            }

fiendie:
Stale

PR #597, opened 2025-10-01
fix small build error

timseidel:
fix build error from non inclusion of shouldDisplayBrewTimer() in main.cpp if OLED_DISPLAY == 0, but FEATURE_BREWSWITCH == 1

How it was:
// Display
#define OLED_DISPLAY    0
...
#define FEATURE_BREWSWITCH      1

resulted in build error because in main.cpp
// Check if brew timer should be shown, only if Display is activated.
#if (FEATURE_BREWSWITCH == 1)
    shouldDisplayBrewTimer(); # -> This is not included, because the display is configured to 0
...

I adjusted it to include the additional check:
... && OLED_DISPLAY != 0 ...

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1669,8 +1669,8 @@ void looppid() {
     updateStandbyTimer();
     handleMachineState();
 
-    // Check if brew timer should be shown
-#if (FEATURE_BREWSWITCH == 1)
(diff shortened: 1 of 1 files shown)


timseidel:
didnt see the whole new version - the PR was unnecessary.

PR #604, opened 2025-12-02
Added initial tests in TSIC measurements

scottcondie24:
This waits for readings to stabilise in an expected range before tightening the max change rate filter.
It should fix issues with bad readings on start, and should filter the worst of the incorrect temperature readings due to delays in read timing.
At full heating power the maximum change rate should be less than 1°C so there could be a lot more optimisation.

### DIFF
--- a/src/hardware/tempsensors/TempSensorTSIC.cpp
+++ b/src/hardware/tempsensors/TempSensorTSIC.cpp
@@ -7,7 +7,8 @@
 #include "TempSensorTSIC.h"
 #include "Logger.h"
 
-#define MAX_CHANGERATE 15
+#define INITIAL_CHANGERATE 200
+#define RUNTIME_CHANGERATE 5
 
 TempSensorTSIC::TempSensorTSIC(const int GPIOPin) {
     // Set pin to receive signal from the TSic 306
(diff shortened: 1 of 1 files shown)


LoQue90:
my temp errors on boot are gone. 
Temp readings while machine is running are looking good.

LoQue90 (approved):
maybe temp readings above 200 could be filtered because there are the errors from the lib.
Then there would be no log warning for these readings.

scottcondie24:
> maybe temp readings above 200 could be filtered because there are the errors from the lib. Then there would be no log warning for these readings.

I'm glad it's working. Can you clarify what change you want? Currently it should still print reading failed and not connected errors if detected. The message level could be changed if you want or the log lines removed entirely

LoQue90:
nevermind, i interpreted the logf wrong.
maybe we could be a bit more specific.
Not the temp itself is out of range, the diff between to values is out of range.

scottcondie24:
> nevermind, i interpreted the logf wrong. maybe we could be a bit more specific. Not the temp itself is out of range, the diff between to values is out of range.

That makes sense. I wasn't sure how often it would send messages if I added logs to say its not matching the criteria yet.

The current code sends a warning if its not within 0 - 180, this could be clarified with a message like "Temperature reading outside 0 - 180°C range" and it could also block 221 and 222 as they log later
It tests if the previous reading was also within 0-180, and that their difference is < 5°C, once that is met then it uses the 5°C in all future temp readings.

fiendie (approved):

PR #605, opened 2025-12-02
ISR fix and reinitialise on freeze

scottcondie24:
timerAlarmWrite(timer, 10000, true); was being called each interrupt, even though it was already set to automatically restart. In rare cases this could corrupt causing the ISR to stop firing, which has been seen on multiple occasions where the heater is stuck on.

A counter has been added to enable checking if it is still active, and reinitialise it if it stops. It also turns the heater off during reinitialisation in case the timer stopped while the heater was on.

### DIFF
--- a/src/isr.h
+++ b/src/isr.h
@@ -11,13 +11,12 @@
 
 extern double pidOutput;
 
-unsigned int isrCounter = 0; // counter for ISR
+unsigned int isrCounter = 0;  // counter for ISR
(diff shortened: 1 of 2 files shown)


scottcondie24:
To test the concept I temporarily added the extern bool steamON into the isr where isrWatchdog only increments if !steamON, then printed the value of isrWatchdog and lastIsrWatchdog to confirm it operated as expected.

fiendie (approved):

PR #607, opened 2025-12-07
Update version of AcaiaArduinoBLE fork to add support for Bookoo scales

fiendie:
Self-explanatory :)


scottcondie24 (approved):
Works well on my old Acaia scales

PR #615, opened 2025-12-28
Refactor Config/ParameterRegistry

fiendie:
- Fix potential dangling pointer in ParameterRegistry
- Add boolean flag for parameters that require a reboot after value change
- Reduce code duplication in Config


scottcondie24 (approved):
It works great! I see a bit of String usage but that shouldn't impact heap enough to worry. The scale samples only gets used during a calibrate or initialise so could also get a reboot flag

PR #619, opened 2026-01-04
Fix scales not failing if disconnected

scottcondie24:
Th hx711 library only sets data pins to input, so they are floating if there is nothing connected which will sometimes pass initialisation.

This remaps the data pin to input_pullup so it will reliably fail if there is a connection issue.


fiendie (approved):

fiendie:
Merged manually due to merge conflicts with VERSION.txt

PR #620, opened 2026-01-04
Flow rate calculation and predictive target weight

scottcondie24:
Using the scales it calculates the flow rate using least squares, a method used to determine the trend line of a set of data. 
Using the flow rate it stops the brew early to end with a weight closer to the desired target. The amount that trickles out varies with different flow rates, so this should get closer than a static offset.
A check for kBrewFinished was added skip long tasks for one cycle if brew has finished, improving performance.


scottcondie24:
removed full screen display changes, flow rate is still visible through mqtt at 5s intervals and 500ms intervals during brew

PR #621, opened 2026-01-21
Added support for wokwi simulator

r3vile:
- Added Wokwi configuration files (`wokwi.toml`, `diagram.json`) for emulator setup.
- Updated `CONTRIBUTING.md` to include Wokwi setup instructions.
- Introduced a new script (`platformio_wokwi.py`) for building firmware and generating Wokwi flasher arguments.
- Added example configuration file (`example_config.json`) for usage with the emulator.
- Implemented WiFi configuration options in `Config.h` and `defaults.h` to support Wokwi's WiFi setup.


scottcondie24:
I needed to change the two override returns in wokwi_flasher_args.py to this to compile on Windows 10. I'm not sure how that would affect other OS. Without posix=False it was returning no slashes

return shlex.split(override, posix=False)

PR #628, opened 2026-05-23
precompile html to reduce load

scottcondie24:
This removes the html_fragments code from staticProcessor by precompiling it in a python script during the build process.
It adds a skip list to the compression script as the header fragment no longer needs to be copied.
It swaps the heap check in load-libs for a simple short wait as its not working how its intended.

These changes help to load the webpage when the ESP is already stressed, leading to less reboots.


scottcondie24:
I fixed a bug in the load-libs due to a conflicting name

irrwisch1:
The preprocessing of the HTML and the fix in the firmware seem like a good idea to me. AsyncWebServer supports chunked streaming directly from LittleFS, so there shouldn't be a problem with the heap.

But the delay in Javascript seems wrong to me. It's not a good idea to work around issues in the backend with a delay in the frontend. If the backend can't keep up because it's short on resources, then the backend should be the one throttling, not the frontend.

I might try these fixes (without the javascript part) on a test board and see if I can reproduce a crash there by stressing the board using curl. With the coredump support that I added in #645 it's quite comfortable to debug these kinds of things.

Also, is there a reason why load-libs.js is not compressed like the other JS files, or is this just an oversight?

scottcondie24:
Yes the delay is not ideal, but the other method being used to throttle also wasn't working and this was an improvement. It's likely the issues caused by the concurrent requests could be fixed elsewhere but I found supplying the files sequentially improved reliability.

I'm curious to see what insights the coredump support gives, the console log does print a lot of detail for crashes.

Good idea on compressing Load-libs, I hadn't considered it

irrwisch1:
I ran this on a test board against upstream master: 60 seconds of load on `/`, four concurrent curl clients. The results are clear, the backend changes really help:

| | requests | answered | crashes |
|---|---:|---:|---:|
| upstream master | 548 | 93 | **5** |
| this PR | 555 | **555** | **0** |

Every response on this PR was the full page, and the board never reset.

The crash on master is an `abort()` in `_fillBufferAndProcessTemplates`, where the substituted page gets inserted into a `std::vector` and `operator new` throws:

```
_fillBufferAndProcessTemplates      (WebResponses.cpp:599)
  -> std::vector<uint8_t>::_M_range_insert
  -> operator new -> __cxa_throw -> std::terminate -> abort()
```

Taking the fragment loading out of `staticProcessor` really does fix it

curl does not run JavaScript, so none of the result above comes from the delay. The backend change carries it on its own, and it also covers the clients that would not see the delay: a second tab, a monitoring script, etc... I would keep the precompilation and drop the delay.

scottcondie24:
> I ran this on a test board against upstream master: 

That's a great result, thanks for testing.

While stressing the code with my dimmer pr, MQTT and Bluetooth scale I found the delay was more reliable than the existing method. I can drop it if it really doesn't make a difference

scottcondie24:
> I would keep the precompilation and drop the delay.

I rebased and removed the delay time. If required it would become a separate PR

irrwisch1:
On a side node: The template processor is not required at all. The only thing it does after this PR is replacing the Version string. The version could be passed through the /parameters endpoint or a seperate /version endpoint. If the template processor is removed the whole html could be just served statically which would probably improve things even further.

Also bundling of js (and css) would make the whole load-libs.js stuff unnecessary and also reduce the load on the ESP32 because loading the frontend then are just a low number of chunked HTTP request instead of multiple requests.

sway:
Also tested on my dev board and prod machine. Works fine and makes things a bit snappier ✅.

irrwisch1:
I'm currently making some improvements the the frontend, based on this PR:

- Bundling of the js/css with esbuild
- removing the template processor and serving the HTML statically and precompressed
- Removing the FontAwesome font and replaced the few glyphs that are used with static SVG (because the font is quite bloated and only a few glyphs are used)
- Converted the web page to a single page application
- optimized /parameters endpoint for the home page to fetch only the parameters that are used. Right now the home page fetches /parameters three times to get all of then, but it only actually uses 3 of these parameters.

Right now I have all these changes in my own fork and are still testing and fixing bugs. But so far it's looking very good, reducing the load on the ESP32 and making the web frontend a bit snappier.
I will eventually put these changes as seperate PRs, but since my changes all build on this PR I'm going to wait until this PR is merged.

scottcondie24:
> On a side node: The template processor is not required at all. 

I couldn't think of a future use for it, but figured that didn't mean it wasn't still useful later. There are likely still a few functions like that one that could be removed

scottcondie24:
> I'm currently making some improvements the the frontend, based on this PR:
> 
> * Bundling of the js/css with esbuild
> * removing the template processor and serving the HTML statically and precompressed
> * Removing the FontAwesome font and replaced the few glyphs that are used with static SVG (because the font is quite bloated and only a few glyphs are used)
> * Converted the web page to a single page application
> * optimized /parameters endpoint for the home page to fetch only the parameters that are used. Right now the home page fetches /parameters three times to get all of then, but it only actually uses 3 of these parameters.
> 
> Right now I have all these changes in my own fork and are still testing and fixing bugs. But so far it's looking very good, reducing the load on the ESP32 and making the web frontend a bit snappier. I will eventually put these changes as seperate PRs, but since my changes all build on this PR I'm going to wait until this PR is merged.

Sounds like some good improvements are coming. I tried to bundle the js/css a while ago but couldn't get it to a polished state. 
FontAwesome was painful before compression, since then I haven't considered what could be optimised there.
The /parameters have had many incremental improvements, it's good to see continued optimisation on it as it's still bloated

scottcondie24:
> Also tested on my dev board and prod machine. Works fine and makes things a bit snappier ✅.

Thanks for testing it

PR #630, opened 2026-06-04
Add pressure to fullscreen brew timer

scottcondie24:
Pressure visible in both horizontal and vertical orientations when pressure sensor is enabled.

Spacing adjusts to fit scales if enabled.

Smaller 15 font used so it's still easily visible but minimises clutter.

Potential improvements to the fullscreen layouts could include using functions like getUTF8Len to right align text. I'm not sure if there is an equivalent for height but that would make smaller units easier to include.


sway:
FWIW I realized that using `getUTF8Len` is not really needed since the text has highly predictable width - `nn.nx` will always have the same pixel width regardless of whether the number is 11.1 or 99.9. I've updated the fullscreen view in my font branch so will also backport your pressure PR into it.

scottcondie24:
> FWIW I realized that using `getUTF8Len` is not really needed since the text has highly predictable width - `nn.nx` will always have the same pixel width regardless of whether the number is 11.1 or 99.9. I've updated the fullscreen view in my font branch so will also backport your pressure PR into it.

Thanks, that's a good point. UTF8 length only varies when there are extended ASCII characters (there are some in languages.h) or fonts that aren't a uniform width, so shouldn't be necessary for this view.

sway:
@scottcondie24 I was playing around with some options for how to best display this. I think with 3 data points there needs to be some visual hierarchy, so I thought of some options for this. Wdyt?

image

LoQue90:
I do like the vertical approach with hirachy. :)

PR #635, opened 2026-07-05
Option: heat while water tank is empty

stan23:
Add a configuration parameter to keep the PID and heating on in case the water tank is reported as empty. The boiler is usually still full of water.

Export the MQTT sensor "~~waterTankFull~~" "waterTankEmpty" because the machine state is not sufficient anymore to detect an empt tank. The state is recognised as ~~dry/wet~~ "problem" by HomeAssistant.

image


fiendie (dismissed):

stan23:
Build problem is caused by https://github.com/platformio/platform-espressif32/issues/1769

